PR #7, opened 2023-07-02
Configure `pg_hba.conf` file

shivaraj-bh:


srid:
Is there a particular use-case (in our projects) for this? 

[The docs](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) also say,

> A default pg_hba.conf file is installed when the data directory is initialized by [initdb](https://www.postgresql.org/docs/current/app-initdb.html).

Is that default not sufficient for our projects (the module already runs initdb)?

shivaraj-bh:
Default is not sufficient as the authentication method for superuser (i.e. `postgres`) is set to `trust` and due to that setting the authentication method of all users using `POSTGRES_INITDB_ARGS` fails saying that it requires password for superuser. There are two options here, set the password for superuser using `POSTGRES_PASSWORD` (a bit hacky) or provide a cleaner API to configure authentication methods with `services-flake`.

nix/postgres.nix
```diff
@@ -44,6 +44,37 @@ in
           default = "./data/${config.name}";
           description = "The DB data directory";
         };
+
+        pgHbaConf =
+          let
+            mkStringOption = lib.mkOption { type = lib.types.str; };
```
srid, line 50: IMO we should inline this, as in future we can add a `description` here as well.

nix/postgres.nix
```diff
@@ -44,6 +44,37 @@ in
...
+            pgHbaConfSubmodules = lib.types.submodule {
```
srid, line 51: The naming here is plural, but the value it points to is singular (one submodule).

nix/postgres.nix
```diff
@@ -44,6 +44,37 @@ in
           default = "./data/${config.name}";
           description = "The DB data directory";
         };
+
+        pgHbaConf =
```
srid, line 48: I'd rename this to `hbaConf`, because the `pg` prefix is redundant (this is already inside `serivices.postgresql`).

nix/postgres.nix
```diff
@@ -257,6 +308,9 @@ in
               # Setup config
               set -x
               cp ${configFile} "$PGDATA/postgresql.conf"
+
+              # Configure authentication
+              echo "${pgHbaConfString}" | ${pkgs.util-linux}/bin/column -t -s $'\t' > "$PGDATA/pg_hba.conf"
```
srid, line 313: Do we really need to create this file separately? Why not put the store path in `settings.hba_file`? See https://github.com/NixOS/nixpkgs/blob/ee5cc38432031b66e7fe395b14235eeb4b2b0d6e/nixos/modules/services/databases/postgresql.nix#L443

srid (approved):
I'm not entirely familiar with this part of postgres, but if it is needed for nammayatri let's do it.

shivaraj-bh:
> I'm not entirely familiar with this part of postgres, but if it is needed for nammayatri let's do it.

I have used this before. It is not very useful if the postgres server is limited to your local machine, but when you want to expose your postgres on a network (for example, to keep the postgres on a separate machine for loadtest) and configure authentication rules for the users this can be a helpful addition.

nix/postgres.nix
shivaraj-bh: I was thinking of using `writeText` for this but wasn't really sure how I can use `column` tool to make the configuration file readable.

srid: You need `pkgs.runCommand` I think (then write to `$out` inside).

shivaraj-bh: I have updated the PR to use `pkgs.runCommand` and I am using `settings.hba_file`. This PR should be good to be merged. I think it will be good to merge after adding a nixos test to verify if the `hba_file` being used by postgres is the same as the one created by `runCommand` in `/nix/store/...-pg-hba.conf`.

shivaraj-bh: I think it is enough to have a single line test to check if `hba_file` path starts with `/nix/store`

nix/postgres.nix
```diff
@@ -77,32 +107,58 @@ in
...
+
+            # Convert the pgHbaConf array to a string
+            hbaConfString = ''
+              # Generated by Nix
+              ${"# TYPE\tDATABASE\tUSER\tADDRESS\tMETHOD\n"}
+              ${lib.concatMapStrings (cnf: "  ${cnf.type}\t${cnf.database}\t${cnf.user}\t${cnf.address}\t${cnf.method}\n") hbaConf}
+            '';
+            hbaDrv = pkgs.runCommand "pg_hba.conf" { } ''
+              echo "${hbaConfString}" | ${pkgs.util-linux}/bin/column -t -s $'\t' > $out
+            '';
```
srid, line 133: I'd put this entire thing an `internal` (and `readOnly`) option alongside `config.hbaConf`. Maybe call it `config.hbaConfFile`. Doing so decouples the hba stuff from the `settings` option thus making it simpler to read.

nix/postgres.nix
```diff
@@ -77,32 +107,58 @@ in
...
+              for an overview of `postgresql.conf`.
+
+              String values will automatically be enclosed in single quotes. Single quotes will be
+              escaped with two single quotes as described by the upstream documentation linked above.
+            '';
+            default = {
+              listen_addresses = config.listen_addresses;
+              port = config.port;
+              unix_socket_directories = lib.mkDefault config.dataDir;
+              hba_file = "${hbaDrv}";
```
srid, line 150: This would then simply become `hba_file = config.hbaConfFile` without the let block.

shivaraj-bh:
@srid I have decoupled the hbaFile creation logic out of settings and also included a nixOS test to check for the usage of `hba_file` by postgres. Looks good to me, should I go ahead and merge?

nix/postgres.nix
```diff
@@ -44,6 +44,67 @@ in
...
+
+            # Convert the pgHbaConf array to a string
+            hbaConfString = ''
+              # Generated by Nix
+              ${"# TYPE\tDATABASE\tUSER\tADDRESS\tMETHOD\n"}
+              ${lib.concatMapStrings (cnf: "  ${cnf.type}\t${cnf.database}\t${cnf.user}\t${cnf.address}\t${cnf.method}\n") hbaConf}
+            '';
+          in
+          lib.mkOption {
+            type = lib.types.raw;
```
srid, line 99: This should be a proper type, `types.path` or `types.package`. raw is too permissive.

srid (approved):
@shivaraj-bh LGTM, feel free to squash merge after that one last change replacing raw type.

PR #12, opened 2023-07-05
util-linux breaks services-flake on macos

shivaraj-bh:


nix/postgres.nix
```diff
@@ -100,9 +100,7 @@ in
             internal = true;
             readOnly = true;
             description = "The `pg_hba.conf` file.";
-            default = pkgs.runCommand "pg_hba.conf" { } ''
-              echo "${hbaConfString}" | ${pkgs.util-linux}/bin/column -t -s $'\t' > $out
```
srid, line 104: This should work on macOS: `lib.getExe pkgs.unix-tools.column`

PR #13, opened 2023-07-05
Multiple services

### LINKED ISSUE #9: Multiple services
cf. "*This also solves the problem of only being able to run one instance at a time.*" of https://github.com/cachix/devenv/issues/75

## Consider this scenario

`passetto` adds a postgresql service. But the user (nammayatri) may add their own postgresql service. Can the two DB coexist? 

## What we can do

Should we adjust the module API to accomodate for this? For eg., instead of:

```nix
services.postgresql.enable = true;
```

do we instead do:

```nix
services.postgresql."some-name".enable = true;
```

And consistently use this API for all services?

## Concrete goal

This should work in nammayatri which has [its own postgresql](https://github.com/nammayatri/nammayatri/blob/55c501f4e79fe1b9d7d8b78b52db3f7d1374616d/Backend/nix/arion-configuration.nix#L11-L14), as well as [a postgres db used for passeto](https://github.com/nammayatri/nammayatri/blob/55c501f4e79fe1b9d7d8b78b52db3f7d1374616d/Backend/nix/arion-configuration.nix#L107-L110).

shivaraj-bh:
Run `, ex` to see two instances of postgres services running on port 5432, 5433 respectively.

Resolves #9 

**Note: This is going to be a breaking change**

Tasks:
- [x] Init with postgres
- [x] Add a inner submodule `config` that defines the behaviour of each service and merge them together on the top-level `config`
- [x] Apply this change in `hello` and `redis` services
- [x] Fix `redis` and `postgres` NixOS tests.
- [x] Add NixOS tests

Example:
## Redis
```nix
# Inside perSystem
{
  # This adds a `self.packages.default`
  process-compose."default" = { config, ... }:
    {
       imports = [
         inputs.services-flake.processComposeModules.default
       ];
       # Starts redis server on port 6379 with data directory in ./data/redis1
       services.redis.redis1.enable = true;
       services.redis.redis2 = {
         enable = true;
         port = 6380;
       };
    };
}
```


nix/postgres.nix
```diff
@@ -236,153 +231,159 @@ in
...
-              failure_threshold = 5;
-            };
-            # https://github.com/F1bonacc1/process-compose#-auto-restart-if-not-healthy
-            availability.restart = "on_failure";
-          };
+        };
       };
+    }));
+  };
+  config = let mergeMapAttrs = f: attrs: lib.mkMerge (lib.mapAttrsToList f attrs); in {
```
shivaraj-bh, line 386: Maybe move `mergeMapAttrs` elsewhere for other services to use it as well? Or would it be a wiser option to just inline this? @srid any thoughts?

srid: Maybe latter: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

nix/postgres.nix
```diff
@@ -236,153 +231,159 @@ in
...
+    settings.processes = mergeMapAttrs (name: cfg: lib.mkIf cfg.enable cfg.outputs) config.services.postgres;
```
srid, line 387: Can't this be written using `imports`?

```nix
{
  settings.processes = {
    imports = lib.mapAttrs (_: cfg: lib.mkIf cfg.enable cfg.outputs) config.service.postgres;
  };
}
```

shivaraj-bh: I get an error complaining about `settings.processes.imports` not being of type `submodule`

srid: We should probably output the entire settings then (`outputs.settings`); making it an import-eable module provides the advantage of letting the module system deal with conflicts.

shivaraj-bh: You mean then we can do something like:
```nix
{
  config.settings.imports = lib.mapAttrsToList (_: cfg: lib.mkIf cfg.enable cfg.outputs) config.services.postgres;
}
```
Correct me if I am wrong, that shouldn't be possible because `imports` option doesn't exist in `process-compose.default.settings`.

srid: My error. I meant the top-level submodule. So, `output` option would set the _whole_ process-compose config (with the expectation that they are normally contain _only_ `processes`).

The you can do,

{
  config = {
    imports = ...;
  };
}
```

srid: Though I wonder if we should create `settings` using `submoduleWith` (to avoid [`shorthandOnlyDefinesConfig`](https://github.com/NixOS/nixpkgs/blob/78386f6b895bb3afb33732dbf8d0eb6dae809194/lib/types.nix#L721-L725) disabling imports) as well ...

shivaraj-bh: Defining either `settings` or `processes` option with `submoduleWith` looks like the right direction, I will try this out.

srid (approved):

PR #17, opened 2023-07-10
readme: Add gif demo of ./example

srid:


srid:
Produced using https://github.com/asciinema/agg

Feel free to replace this with a better/improved version latter.

PR #24, opened 2023-07-19
add baseDataDir to redis and postgres

### LINKED ISSUE #23: add baseDataDir which could be shared between services
I'd like to override default base `dataDir` for all services, without having to re-state service name in each override

I'd like to do this:

```nix
let
  baseDataDir = "./.data";
in {
  services.redis."redis-1" = {
    enable = true;
    inherit baseDataDir;
  };
  services.postgres."pg-1" = {
    enable = true;
    inherit baseDataDir;
  };
}
```

instead of:

```nix
let
  baseDataDir = "./.data";
in {
  services.redis."redis-1" = {
    enable = true;
    dataDir = "${baseDataDir}/redis-1";
  };
  services.postgres."pg-1" = {
    enable = true;
    dataDir = "${baseDataDir}/pg-1";
  };
}
```

while preserving the original data dir structure

see possible implementation [here](https://github.com/rhinofi/services-flake/commit/e35c04f9203755877bce446edb51f7db94824c9c)

adrian-gierakowski:
closes #23

### DIFF
--- a/nix/postgres.nix
+++ b/nix/postgres.nix
@@ -27,10 +27,16 @@ in
       '';
     };
 
+    baseDataDir = lib.mkOption {
+      type = types.str;
+      default = "./data";
+      description = "Base data dir. By default, each instance will store it's data in $baseDataDir/$name";
+    };
+
     dataDir = lib.mkOption {
-      type = lib.types.str;
-      default = "./data/${name}";
-      description = "The DB data directory";
+      type = types.str;
+      default = "${config.baseDataDir}/${name}";
+      description = "Instance data directory.";
     };
 
     hbaConf =
--- a/nix/redis.nix
+++ b/nix/redis.nix
@@ -9,10 +9,16 @@ in
 
     package = lib.mkPackageOption pkgs "redis" { };
 
+    baseDataDir = lib.mkOption {
+      type = types.str;
+      default = "./data";
+      description = "Base data dir. By default, each instance will store it's data in $baseDataDir/$name";
+    };
+
     dataDir = lib.mkOption {
       type = types.str;
-      default = "./data/${name}";
-      description = "The redis data directory";
+      default = "${config.baseDataDir}/${name}";
+      description = "Instance data directory.";
     };
 
     bind = lib.mkOption {


shivaraj-bh:
Screenshot 2023-07-19 at 7 12 05 PM
Verified that the NixOS tests pass! Until we have a CI setup, this is a convenient way to test PRs.

srid (commented):
You still have to specify `baseDataDir` for each of the service *type* (if not each of their instances). I'm wondering if there is a way to specify this option once for all service types. 

Something like `services.defaults.baseDataDir`?

adrian-gierakowski:
> Verified that the NixOS tests pass! Until we have a CI setup, this is a convenient way to test PRs.

thanks @shivaraj-bh ! I'll run this locally next time 

@srid 

> You still have to specify baseDataDir for each of the service type (if not each of their instances). I'm wondering if there is a way to specify this option once for all service types.

> Something like services.defaults.baseDataDir?

I was going to suggest it next :)

Do you think we should simply use `services.defaults` keeping the current structure, or should we nest service groups under `services.groups' or `services.instances`?

shivaraj-bh:
IMO we could keep the current structure as proposed in the PR and have `services.default`. `baseDataDir` in each service gets its value from `services.default.baseDataDir` or if null, it is set to `./data`. The user will still be able to explicitly specify the data dir for a specific service.

srid:
Normally I'd go with `services.groups` (to keep things separate) but `services.groups.postgres."pg1".enable` is already getting verbose.

Let's take a step back. What if we relied on an environment variable?

```nix
{ 
  process-compose."default" = {
    environment.SERVICES_DATA_DIR = "./.data"; # Override the default (./data)
    services.postgres."pg1".enable = true;
  };
}
```

adrian-gierakowski:
> Normally I'd go with `services.groups` (to keep things separate) but `services.groups.postgres."pg1".enable` is already getting verbose.
> 
> Let's take a step back. What if we relied on an environment variable?
> 
> ```nix
> { 
>   process-compose."default" = {
>     environment.SERVICES_DATA_DIR = "./.data"; # Override the default (./data)
>     services.postgres."pg1".enable = true;
>   };
> }
> ```

yeah, that could (partially) work, but personally I wouldn't use it

I wrote partially since `environment` variables set this way are not available to shell snippet commands since PC substitutes env vars in the config file using the env in which PC was ran, before setting env vars defined in the config file

so in the following example the `SERVICES_DATA_DIR` would be empty in `env-test-shell-snippet`, but set in the `run.sh` script 


```yaml
environment:
  - SERVICES_DATA_DIR=./.data
processes:
  env-test-shell-snippet:
    command: "while true; do sleep 1; echo SERVICES_DATA_DIR: $SERVICES_DATA_DIR; done"
  env-test-script:
    command: ./run.sh
shell:
  shell_argument: -c
  shell_command: bash
```

srid:
We are so far using wrapper scripts (`writeShellScript`, etc) in services, and have't come across a situation where inline commands are needed or desired.

If we do go with the environment approach, there is a problem: how do we reference it in option defaults like here? 

https://github.com/juspay/services-flake/blob/321ef32a28ed0e59160cf130c3a7c55e3b979e85/nix/postgres.nix#L142C11-L143C1

---

There is another larger problem with this whole data-directory business (environment or otherwise). It is usually a relative path, but there is no guarantee as to what it is relative to. Presently, it is relative to $PWD, and that can cause surprises: #26. 

I feel like to we should think about all of this a bit and come up with an ideal solution.

srid:
With https://github.com/juspay/services-flake/issues/23#issuecomment-1643007562 now - is there any reason for this PR, @adrian-gierakowski ?

srid:
Closing per https://github.com/juspay/services-flake/issues/23#issuecomment-1643835355

PR #27, opened 2023-07-19
Use `just` for deterministic `dataDir` in Nix shell

### LINKED ISSUE #26: Dealing with non-root $PWD
```
cd ./example
mkdir tmp
cd ./tmp
nix run
```

This creates the data diretory in `./example/tmp/data` (instead of `./example/data` as one would expect -- ie. in flake root). What is the best way to address this? In nammayatri we use `flake-root` + `mission-control` module to side-step the problem. 

One possible solution here is to have services-flake *require* the use of `flake-root` module, and have the data directory use the flake root directory determined by the latter module.

srid:
First, this PR switches the dev flake from mission-control to [just](https://github.com/casey/just) similar to https://github.com/srid/haskell-template/pull/111

Then, we introduce `justfile` under the example flake. Now if you do,

```sh
cd ./example
mkdir foo 
cd ./foo
nix run
```

The data directory will be created under `./example/foo` which is unexpected. Whereas, the following would always create it in `./example`:

```sh
cd ./example
mkdir foo 
cd ./foo
just
```

Unless `[no-cd]` is used, [just](https://github.com/casey/just#recipe-parameters) will always cd to the directory where the `justfile` lives. 

The README documents this, as a sufficient solution such that it resolves #26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,10 @@ TODO
 - [x] Redis
 - [ ] ...
 
+## A note on process working directory
+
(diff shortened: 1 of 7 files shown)


srid:
Btw, shell autocompletion for `just` works in nix shell (on macos, one has to use `bash`)

PR #31, opened 2023-07-20
feat: add lib to flake outputs exposing multiService func

### LINKED ISSUE #25: expose multiService function on lib output
I found [multiService](https://github.com/juspay/services-flake/blob/321ef32a28ed0e59160cf130c3a7c55e3b979e85/nix/default.nix#L8-L33) useful for defining my local services, so ended up copying the code.

adrian-gierakowski:
closes #25


flake.nix
```diff
@@ -7,6 +7,10 @@
       path = builtins.path { path = ./example; filter = path: _: baseNameOf path == "flake.nix"; };
     };
 
+    # NOTE: this is a lib factory, as it takes: { pkgs, lib }
+    # so before using we need to pass it to pkgs.callPackage
+    lib = import ./nix/lib.nix;
```
adrian-gierakowski, line 12: not sure if it's common for the `lib` output to be a function but `nix flake check` didn't complain

for this to be a function we'd have add nixpkgs as flake input

srid: We could use `inputs.nixpkgs.lib` but since the submodule below passes `pkgs` as well, as specialArgs, we necessarily need some perSystem context. @roberth is there an idiomatic way here?

adrian-gierakowski: I wonder if it would be better to expose this via an overlay:

```nix
final: _: {
  services-flake = {
    lib = import ./nix/lib.nix { pkgs = final; };
  };
}
```

or if you want to give it a bit more thought, we could not expose it at all and I'd import the file directly via:

```nix
import "${services-flake}/nix/lib.nix" { inherit pkgs; }
```

adrian-gierakowski: Cool, I’ll update the PR accordingly

shivaraj-bh: Thanks! I will merge it once you make the change.

roberth, line 12:
```suggestion
    lib = {
      withPkgs = import ./nix/lib.nix;
    };
```

This leaves room for functions that don't depend on `pkgs`.
Accepting a `lib` attribute in the parameter is implied.

roberth, line 12:
```suggestion
    lib = {
      /*
        Type:
          { lib, pkgs } -> { ... }

        A library of system-specific functions.
      */
      withPkgs = import ./nix/lib.nix;
    };
```

This leaves room for functions that don't depend on `pkgs`.
Accepting a `lib` attribute in the parameter is implied.

roberth (commented):
Some ideas. It looks like your lib doesn't need a `pkgs`.

nix/lib.nix
```diff
@@ -0,0 +1,33 @@
...
+  multiService = mod:
+    let
+      # Derive name from filename
+      name = lib.pipe mod [
+        builtins.baseNameOf
+        (lib.strings.splitString ".")
+        builtins.head
+      ];
+    in
+    { config, ... }: {
```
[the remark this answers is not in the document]
srid: Ya, let's go with this approach. 

But I don't believe you need the `{ lib }` parameter either (you can use `pkgs.lib` instead and move the `let` block inside the module block). Which means this could simply be exposed as `lib.multiService`.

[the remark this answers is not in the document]
roberth: `pkgs.lib` does require configuring and evaluating `pkgs`, so in order to have a more consistent pattern across the ecosystem, it is beneficial to accept `lib` separately. If it doesn't matter in your case, take `lib ? pkgs.lib`.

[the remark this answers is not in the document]
roberth: > Btw. is the lib flake output expected to be `{ lib } -> { ... }` or just `{ ... }`?

It doesn't come with expectations, but you could do `{ ... }` if you take `lib` from the flake inputs.
Here flakes is an example of a framework that does too much, even if it's just fairly simple dependency injection. It's there, so we're expected to use it, even if it's not quite the right thing. Maybe it should be `{ ... } // { withLib = ... }` if you want to both satisfy all expectations and make the interface fully general.
Could be over-engineered or just right. I'll leave that up to you.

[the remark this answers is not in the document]
srid: > `pkgs.lib` does require configuring and evaluating `pkgs`, so in order to have a more consistent pattern across the ecosystem, it is beneficial to accept `lib` separately.

makes sense

srid:
Anything to be done here? @adrian-gierakowski

adrian-gierakowski:
> Anything to be done here? @adrian-gierakowski

this should now be ready to merge

I've removed the need for injecting `pkgs` and `lib` (as you suggested in https://github.com/juspay/services-flake/pull/31#discussion_r1273572257) so now the flake's `lib` output is simply an `attrset`

PR #34, opened 2023-07-21
postgres: add defaultSettings internal option

### LINKED ISSUE #30: postgres: provide a way to add to default settings instead of simply overriding all of them
the following defaults: 

https://github.com/juspay/services-flake/blob/f3a7ea9ec9166f8b97e0f303668d3e0608835e2b/nix/postgres.nix#L139-L143

are needed for the setup to function correctly and would rarely need to be overriden

a user however would often want to add more settings

in my case I'd like to add `shared_preload_libraries = 'timescaledb'`

we could add `extraSettings` which would be merged with `settings` or move current defaults to `defaultSettings` and leave `settings` empty, for the user to set

shivaraj-bh:
closes #30

### DIFF
--- a/nix/postgres.nix
+++ b/nix/postgres.nix
@@ -124,6 +124,22 @@ in
       '';
     };
 
+    defaultSettings =
+      lib.mkOption {
+        type = lib.types.raw;
+        internal = true;
+        readonly = true;
+        description = ''
+          Default configuration for `postgresql.conf`. `settings` can override these values.
+        '';
+        default = {
+          listen_addresses = config.listen_addresses;
+          port = config.port;
+          unix_socket_directories = lib.mkDefault config.dataDir;
+          hba_file = "${config.hbaConfFile}";
+        };
+      };
+
     settings =
       lib.mkOption {
         type = with lib.types; attrsOf (oneOf [ bool float int str ]);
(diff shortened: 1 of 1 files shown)


nix/postgres.nix
```diff
@@ -136,6 +136,22 @@ in
...
+            type = lib.types.raw;
+            internal = true;
+            readonly = true;
+            description = ''
+              Default configuration for `postgresql.conf`. `settings` can override these values.
+            '';
+            default = {
+              listen_addresses = config.listen_addresses;
+              port = config.port;
+              unix_socket_directories = lib.mkDefault config.dataDir;
```
srid, line 150: Is `lib.mkDefault` necessary here?

nix/postgres.nix
```diff
@@ -136,6 +136,22 @@ in
           '';
         };
 
+        defaultSettings =
+          lib.mkOption {
+            type = lib.types.raw;
```
srid, line 141: Shouldn't this be `attrsOf (oneOf [ bool float int str ]);`?

srid (approved):
LGTM (aside from possibly the `type`)

nix/postgres.nix
shivaraj-bh: removed it! We don't care much about the priority of the default here anyway.

nix/postgres.nix
shivaraj-bh: Since it was `internal` I thought it's fine to be raw, but doesn't hurt to be more specific.

PR #35, opened 2023-07-26
Add Redis cluster service

### LINKED ISSUE #15: Add redis cluster service

shivaraj-bh:
Closes #15 

`create-cluster` shell script provided by `redis` includes a sub-command to clean all the nodes, which is just removing all the files from the data directory. We can ignore that and maybe just add it under "How to" section in doc.

Example:

```
# Inside perSystem
{
  # This adds a `self.packages.default`
  process-compose."default" = { config, ... }:
    {
       imports = [
         inputs.services-flake.processComposeModules.default
       ];
       # Starts redis cluster service with nodes on 30001, 30002, 30004, 30005 and 30006 ports respectively.
       services.redis-cluster."cluster1".enable = true;
    };
}
```

TODO:
- [x] Add NixOS test

### DIFF
--- a/nix/postgres.nix
+++ b/nix/postgres.nix
@@ -343,7 +343,7 @@ in
 
                 initdbArgs =
                   config.initdbArgs
-                  ++ (lib.optionals (config.superuser != null) ["-U" config.superuser]);
+                  ++ (lib.optionals (config.superuser != null) [ "-U" config.superuser ]);
 
                 setupScript = pkgs.writeShellScriptBin "setup-postgres" ''
                   set -euo pipefail
--- a/nix/redis-cluster.nix
+++ b/nix/redis-cluster.nix
@@ -14,16 +14,28 @@ in
       description = "The redis-cluster data directory (common for all nodes).";
     };
 
-    port = lib.mkOption {
-      type = types.port;
-      default = 30001;
-      description = ''
-        The TCP port to accept connections.
-
-        Next node will use the next port and so on.
-
-        If port is set to `0`, redis will not listen on a TCP socket.
-      '';
+    nodes = lib.mkOption {
+      type = types.listOf (types.submodule {
+        options = {
+          port = lib.mkOption {
+            type = types.int;
+            description = "The TCP port to accept connections. If port is set to `0`, redis will not listen on a TCP socket.";
+          };
+          extraConfig = lib.mkOption {
+            type = types.lines;
+            description = "Extra configuration for this node. To be appended to `redis.conf`.";
+            default = "";
+          };
+        };
+      });
+      default = [
+        { port = 30001; }
+        { port = 30002; }
+        { port = 30003; }
+        { port = 30004; }
+        { port = 30005; }
+        { port = 30006; }
+      ];
     };
 
     bind = lib.mkOption {
@@ -48,32 +60,11 @@ in
       '';
     };
 
-    nodes = lib.mkOption {
-      type = types.int;
-      default = 6;
-      description = ''
-        Number of nodes in the cluster.
-      '';
-    };
-
     replicas = lib.mkOption {
       type = types.int;
       default = 1;
       description = ''
         Number of replicas per Master node.
-      '';
-    };
-
-    extraConfigs = lib.mkOption {
-      type = types.attrsOf types.lines;
-      default = { };
-      description = "Attrset map of node (identified by port number) to the respective `redis.conf`.";
-      example = ''
-        {
-          {
-            30001 = "port 30001";
-          };
-        }
       '';
     };
 
@@ -83,11 +74,9 @@ in
       readOnly = true;
       default =
         let
-          hosts = lib.genList (x: "${config.bind}:${builtins.toString(config.port + x)}") config.nodes;
-          ports = lib.genList (x: "${builtins.toString(config.port + x)}") config.nodes;
-          healthyNodes = lib.genAttrs ports (port: { condition = "process_healthy"; });
-          node = port:
+          mkNodeProcess = nodeConfig:
             let
+              port = builtins.toString nodeConfig.port;
               redisConfig = pkgs.writeText "redis.conf" ''
                 port ${port}
                 cluster-enabled yes
@@ -98,7 +87,7 @@ in
                 dbfilename "dump-${port}.rdb"
 
                 ${lib.optionalString (config.bind != null) "bind ${config.bind}"}
-                ${lib.optionalString (builtins.hasAttr port config.extraConfigs) "${config.extraConfigs.${port}}"}
+                ${nodeConfig.extraConfig}
               '';
 
               startScript = pkgs.writeShellScriptBin "start-redis" ''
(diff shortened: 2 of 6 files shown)


nix/postgres.nix
```diff
@@ -343,7 +343,7 @@ in
 
                 initdbArgs =
                   config.initdbArgs
-                  ++ (lib.optionals (config.superuser != null) ["-U" config.superuser]);
+                  ++ (lib.optionals (config.superuser != null) [ "-U" config.superuser ]);
```
srid, line 346: Wasn't this caught my nixci (the treefmt-nix check)?

nix/redis-cluster.nix
```diff
@@ -0,0 +1,145 @@
...
+      default = 1;
+      description = ''
+        Number of replicas per Master node.
+      '';
+    };
+
+    extraConfigs = lib.mkOption {
+      type = types.attrsOf types.lines;
+      default = { };
+      description = "Attrset map of node (identified by port number) to the respective `redis.conf`.";
```
srid, line 70: This seems rather unconventional - to have port number as attrset keys? Besides, if someone changes `port` down the line, the whole attrset would get invalidated, no? I may be missing something, but why not something like:

```nix
{
  redis-cluster.nodes = {
    node1 = { port = 30001; extraConfig = "..."; };
    node2 = { port = 30002; extraConfig = "..."; };
  };
}
```

.. which seems rather straightforward to read.

nix/redis-cluster.nix
```diff
@@ -0,0 +1,145 @@
...
+              };
+
+              # https://github.com/F1bonacc1/process-compose#-auto-restart-if-not-healthy
+              availability.restart = "on_failure";
+            };
+        in
+        {
+          processes = lib.genAttrs ports (port: node port) //
+            {
+              "cluster-create" = {
```
srid, line 136: Shouldn't this be `"${name}-cluster-create"` following the naming convention elsewhere?

Eg.:

https://github.com/juspay/services-flake/blob/d30b8afd55d1a5e5958d54c8e6fde2cad39faf48/nix/postgres.nix#L272

nix/redis-cluster.nix
```diff
@@ -0,0 +1,145 @@
...
+    outputs.settings = lib.mkOption {
+      type = types.deferredModule;
+      internal = true;
+      readOnly = true;
+      default =
+        let
+          hosts = lib.genList (x: "${config.bind}:${builtins.toString(config.port + x)}") config.nodes;
+          ports = lib.genList (x: "${builtins.toString(config.port + x)}") config.nodes;
+          healthyNodes = lib.genAttrs ports (port: { condition = "process_healthy"; });
+          node = port:
```
srid, line 89: Function names should be descriptive. Wouldn't something like `mkNodeProcess` be clearer?

While reading the code below I was stuck the expression `(port: node port)` and it wasn't immediately clear what it was doing, until I had to read implementation of the function being called.

nix/redis-cluster.nix
```diff
@@ -0,0 +1,145 @@
...
+                success_threshold = 1;
+                failure_threshold = 5;
+              };
+
+              # https://github.com/F1bonacc1/process-compose#-auto-restart-if-not-healthy
+              availability.restart = "on_failure";
+            };
+        in
+        {
+          processes = lib.genAttrs ports (port: node port) //
```
srid, line 134: Shouldn't this use `${name}-node-${port}` as the key, rather than `${port}` per se, following the naming convention elsewhere?

nix/postgres.nix
shivaraj-bh: I don't think it did, or maybe it went unnoticed

nix/redis-cluster.nix
shivaraj-bh: This makes more sense. I was trying to replicate the create cluster script provided by redis that allocates nodes incrementally on ports one after another, that's why I ended up doing this as init. Will change the behaviour as per the API you mentioned, which will be cleaner.

nix/redis-cluster.nix
shivaraj-bh: Will change the naming. I was bootstrapping a cluster that works on one node and then I generalised it to multiple nodes, but forgot to change the function name.

nix/redis-cluster.nix
shivaraj-bh: Yes! It needs to follow the multiple-services naming convention.

nix/redis-cluster.nix
```diff
@@ -0,0 +1,138 @@
...
+                  failure_threshold = 5;
+                };
+
+                # https://github.com/F1bonacc1/process-compose#-auto-restart-if-not-healthy
+                availability.restart = "on_failure";
+              };
+            };
+          hosts = builtins.map (node: "${config.bind}:${builtins.toString node.port}") config.nodes;
+          healthyNodes = builtins.map (node: { "${name}-node-${builtins.toString node.port}".condition = "process_healthy"; }) config.nodes;
+          createClusterProcess = {
```
srid, line 125: Since this is not a function, a noun is more appropriate than a verb.

nix/redis-cluster.nix
```diff
@@ -0,0 +1,138 @@
...
+            "${name}-cluster-create" = {
+              depends_on = lib.mkMerge healthyNodes;
+              command = "${config.package}/bin/redis-cli --cluster create ${lib.concatStringsSep " " hosts} --cluster-replicas ${builtins.toString config.replicas} --cluster-yes";
+            };
+          };
+          processesList = (builtins.map mkNodeProcess config.nodes) ++ [ createClusterProcess ];
+        in
+        {
+          processes = lib.mkMerge processesList;
```
srid, line 134: How does this behave when there is a key conflict (by accident/bug)? How does `mkMerge` compare to using `imports`?

srid: Why not inline it simply?

```nix
{
  processes = lib.mapAttrs mkNodeProcess (listToAttrs (_ config.nodes)) // {
    "${name}-cluster-create" = clusterProcess;
  };
}
```

shivaraj-bh: > How does this behave when there is a key conflict (by accident/bug)?

Case 1: 
```
{
  config.nodes = [
    { port = 30001; }
    { port = 30002; }
    { port = 30001; }
  ];
}
```
Case 2: 
```
{
  config.nodes = [
    { port = 30001; }
    { port = 30002; }
    { port = 30001; extraConfig = ''maxmemory 2mb''; }
  ];
}
```
In `Case 1` `mkMerge` is considering the first element of the `config.nodes` to resolve the conflict whereas in `Case 2` it will error out saying that "The option `perSystem.aarch64-darwin.process-compose.default.settings.processes.c1-node-30002.command' has conflicting definition values".

shivaraj-bh: > Why not inline it simply?
```nix
{
  processes = lib.mapAttrs mkNodeProcess (listToAttrs (_ config.nodes)) // {
    "${name}-cluster-create" = clusterProcess;
  };
}
```

This can be done, but the problem will be that in the `Case 2` mentioned above it will never throw an error while trying to `nix run` and node `{ port = 30001; extraConfig = ''maxmemory 2mb''; }` will override `{ port = 30001; }`

shivaraj-bh: Decided to go with `attrsOf submodule` instead of `listOf submodule` for `nodes`, as suggested by @srid

shivaraj-bh:
Screenshot 2023-08-02 at 9 57 06 AM

PR #37, opened 2023-08-08
Add elasticsearch service

### LINKED ISSUE #36: Add elasticsearch service
Reference: https://github.com/cachix/devenv/blob/main/src/modules/services/elasticsearch.nix

shivaraj-bh:
Resolves #36 
- [x] Add test

### DIFF
--- a/nix/elasticsearch.nix
+++ b/nix/elasticsearch.nix
@@ -14,6 +14,18 @@ let
 in
 {
   options = {
+    description = ''
+      Configure elasticsearch. This will start a single-node cluster by default.
+      Note: 
+      To use elastic search you will need the following inside your `perSystem`, if not already configured:
+      { 
+        _module.args.pkgs = import inputs.nixpkgs {
+            inherit system;
+            # Required for elastic search
+            config.allowUnfree = true;
+          };
+      } 
+    '';
     enable = lib.mkEnableOption name;
 
     package = lib.mkPackageOption pkgs "elasticsearch7" { };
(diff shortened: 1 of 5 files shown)


nix/elasticsearch.nix
```diff
@@ -0,0 +1,196 @@
...
+{
+  options = {
+    enable = lib.mkEnableOption name;
+
+    package = lib.mkPackageOption pkgs "elasticsearch7" { };
+
+    dataDir = lib.mkOption {
+      type = types.str;
+      # TODO: elasticsearch requires absolute path, is there a way to avoid this?
+      default = "$PWD/data/${name}";
```
shivaraj-bh, line 24: I am not sure if this is the right way to go about this, but it works.

srid: Why not compute absolute path in the script?

```bash
 export ES_HOME="realpath ${config.dataDir}"
```

nix/elasticsearch.nix
```diff
@@ -0,0 +1,196 @@
+# Based on https://github.com/cachix/devenv/blob/main/src/modules/services/elasticsearch.nix
+# Note: 
+#   To use elastic search you will need the following inside your `perSystem`, if not already configured:
+#   { 
+#     _module.args.pkgs = import inputs.nixpkgs {
+#        inherit system;
+#        # Required for elastic search
+#        config.allowUnfree = true;
+#      };
+#   }
```
srid, line 10: This could be put in submodule's `description`?

(nix-browser can render it ultimately, so can flake.parts website)

shivaraj-bh:
Will be adding native test to this PR once https://github.com/juspay/services-flake/pull/38 is merged

srid (approved):

PR #38, opened 2023-08-08
Process as test (VM -> Native test)

### LINKED ISSUE #32: Replace vm tests with native tests
Unlike NixOS services, process-compose programs created by services-flake do not mutate outside of `dataDir`, so we gain very little from the isolation provided by VMs.

Let's switch to running these tests natively, which gets us macOS support (resolving #14), in addition to allowing us to use Github Actions for CI (#11).

## Implementation idea

Use process-compose itself to run tests. process-compose running another process-compose, with readiness checks triggering the test process. We can use bash, or better nushell, to write the tests in. Nushell, because it can easily process the JSON from process-compose API.

shivaraj-bh:
Resolves #32 

Run native test
```bash
just test
```

TODO
- [x] VM -> Native test
  - [x] Postgres
  - [x] Redis
  - [x] Redis cluster

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,17 @@
+name: "CI"
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@main
+      - name: Native test
+        id: test
+        run: |
+          nix flake check test/ --override-input services-flake . -L
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install Nix
         uses: DeterminateSystems/nix-installer-action@main
-      - name: Native test
-        id: test
-        run: |
-          nix flake check test/ --override-input services-flake . -L
+      - uses: yaxitech/nix-install-pkgs-action@v3
+        with:
+          packages: "nixpkgs#nixci"
+      - run: nixci
--- a/example/flake.nix
+++ b/example/flake.nix
@@ -53,6 +53,17 @@
                 command = pkgs.pgweb;
                 depends_on."pg1".condition = "process_healthy";
               };
+            settings.processes.test = {
+              disabled = true;
+              command = pkgs.writeShellApplication {
+                name = "pg1-test";
+                runtimeInputs = [ config.services.postgres.pg1.package ];
+                text = ''
+                  echo 'SELECT version();' | psql -h 127.0.0.1 ${dbName}
+                '';
+              };
+              depends_on."pg1".condition = "process_healthy";
+            };
           };
 
         devShells.default = pkgs.mkShell {
--- a/example/flake.nix
+++ b/example/flake.nix
@@ -4,7 +4,7 @@
     nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
     systems.url = "github:nix-systems/default";
-    process-compose-flake.url = "github:Platonic-Systems/process-compose-flake";
+    process-compose-flake.url = "github:shivaraj-bh/process-compose-flake/merge-yaml-in-nix";
     services-flake.url = "github:juspay/services-flake";
 
     northwind.url = "github:pthom/northwind_psql";
@@ -54,7 +54,6 @@
                 depends_on."pg1".condition = "process_healthy";
               };
             settings.processes.test = {
-              disabled = true;
               command = pkgs.writeShellApplication {
                 name = "pg1-test";
                 runtimeInputs = [ config.services.postgres.pg1.package ];
--- a/example/flake.nix
+++ b/example/flake.nix
@@ -1,5 +1,5 @@
 {
-  description = "A demo of sqlite-web";
+  description = "A demo of sqlite-web and multiple postgres services";
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
--- a/example/flake.nix
+++ b/example/flake.nix
@@ -4,7 +4,7 @@
     nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
     systems.url = "github:nix-systems/default";
-    process-compose-flake.url = "github:shivaraj-bh/process-compose-flake/merge-yaml-in-nix";
+    process-compose-flake.url = "github:Platonic-Systems/process-compose-flake";
     services-flake.url = "github:juspay/services-flake";
 
     northwind.url = "github:pthom/northwind_psql";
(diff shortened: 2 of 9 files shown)


srid:
Could you add a Github Actions workflow for running the new tests?

shivaraj-bh:
Another interesting point to consider here would be to only display the test logs, but filtering out all the other processes' logs might cause problems with visibility because if they fail we won't know why. Essentially we need a way to display logs of a given process if the process ended with a non-zero exit code.

adrian-gierakowski:
> Another interesting point to consider here would be to only display the test logs

I’m currently doing this as follows

```nix
''
  ${pc-warapper} >/dev/null 2>&1 &
  pc_pid=$!

  ${pc-warapper} process logs test -f &
  test_logs_pid=$!

  wait $pc_pid
  kill $test_logs_pid
''
``` 

On first line you could redirect to tmp file and display last X lines if exit code is non-zero.

Note that displaying only the logs of the test process will get easier once [this](https://github.com/F1bonacc1/process-compose/issues/83) is addressed. Logs for all processes could then be collected using `log_location` at the top level of the config file.

shivaraj-bh:
> On first line you could redirect to tmp file and display last X lines if exit code is non-zero.

Thanks! Will give it a shot.

shivaraj-bh:
CI works 🥳

srid:
On macOS (Nix 2.17), I get:

```
❯ just test
nix flake check test/ --override-input services-flake . -L
warning: not writing modified lock file of flake 'git+file:///Users/srid/code/services-flake?dir=test':
• Updated input 'services-flake':
    'github:juspay/services-flake/f171932d130c3ce5184b69144012a225b2e28868' (2023-06-19)
  → 'git+file:///Users/srid/code/services-flake?ref=refs/heads/process-as-test&rev=9928f03781cba69263b83239c6c7bd26d6a1ef97' (2023-08-10)
error: builder for '/nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv' failed with exit code 1
error: build of '/nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv', '/nix/store/d9ssfys8nawjym0cgm26g6m5czamhzpa-redis-cluster-test.drv' failed
error: Recipe `test` failed on line 14 with exit code 1
```

This is odd,

```
❯ nix log /nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '/nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv^*'
```

Tests run on NixOS though.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,17 @@
...
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@main
+      - name: Native test
+        id: test
+        run: |
+          nix flake check test/ --override-input services-flake . -L
```
srid, line 17:
```suggestion
      - uses: yaxitech/nix-install-pkgs-action@v3
        with:
          packages: "nixpkgs#just"
      - run: just test
```

nix/redis-cluster_test.nix
```diff
@@ -1,16 +1,23 @@
...
+    in
+    {
+      command = pkgs.writeShellApplication {
+        text = ''
+          ${cfg.package}/bin/redis-cli -p 30001 ping | ${pkgs.gnugrep}/bin/grep -q "PONG"
+          ${cfg.package}/bin/redis-cli -p 30002 ping | ${pkgs.gnugrep}/bin/grep -q "PONG"
+          ${cfg.package}/bin/redis-cli -p 30003 ping | ${pkgs.gnugrep}/bin/grep -q "PONG"
+          ${cfg.package}/bin/redis-cli -p 30004 ping | ${pkgs.gnugrep}/bin/grep -q "PONG"
+          ${cfg.package}/bin/redis-cli -p 30005 ping | ${pkgs.gnugrep}/bin/grep -q "PONG"
+          ${cfg.package}/bin/redis-cli -p 30006 ping | ${pkgs.gnugrep}/bin/grep -q "PONG"
```
srid, line 16: You can avoid the `${foo}/bin` here by passing these packages to `runtimeInputs` argument of `writeShellApplication`.

srid: Same in postgres test.

.github/workflows/ci.yaml
srid: By the way, why can't we just use `nixci` here?

shivaraj-bh:
> On macOS (Nix 2.17), I get:
> 
> ```
> ❯ just test
> nix flake check test/ --override-input services-flake . -L
> warning: not writing modified lock file of flake 'git+file:///Users/srid/code/services-flake?dir=test':
> • Updated input 'services-flake':
>     'github:juspay/services-flake/f171932d130c3ce5184b69144012a225b2e28868' (2023-06-19)
>   → 'git+file:///Users/srid/code/services-flake?ref=refs/heads/process-as-test&rev=9928f03781cba69263b83239c6c7bd26d6a1ef97' (2023-08-10)
> error: builder for '/nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv' failed with exit code 1
> error: build of '/nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv', '/nix/store/d9ssfys8nawjym0cgm26g6m5czamhzpa-redis-cluster-test.drv' failed
> error: Recipe `test` failed on line 14 with exit code 1
> ```
> 
> This is odd,
> 
> ```
> ❯ nix log /nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv
> warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '/nix/store/c53s53w0xis4ak8jaw1df8cninh92d47-redis-test.drv^*'
> ```
> 
> Tests run on NixOS though.

This also happens on my macOS with `Nix 2.13.3`. Additionally I also tried running the example flake with `just ex` after updating `example/flake.lock` to use the latest PC and the example app doesn't run at all.

I tried on different macOS (with `Nix 2.15.1`) where PC was never ran before and it seems to work fine there. This is weird, have to debug this.

By MacOS I mean `aarch64-darwin` in both the devices mentioned above.

shivaraj-bh:
@srid Now the builds are working on mac, the error wasn't conclusive but it looks like having each PC package run on different port makes it work.

srid:
> A nicer implementation will be to disable swagger endpoint so that we won't have this conflict.

Yea, I was thinking of an upstream contribution to that end. Or we can set port to `0` by default (in process-compose-flake), forcing the user to set it explicitly if they want it.

`just test` runs fine for me now.

srid (approved):
LGTM once process-compose-flake is updated.

example/flake.nix
```diff
@@ -53,15 +53,6 @@
...
               };
-
-            # Set this attribute and get NixOS VM tests, as a flake check, for free.
-            testScript = ''
-              # FIXME: pgweb is still pending, but only in VM tests for some reason.
-              process_compose.wait_until(lambda procs:
-                procs["pg1"]["status"] == "Running"
-              )
-              machine.succeed("echo 'SELECT version();' | ${config.services.postgres.pg1.package}/bin/psql -h 127.0.0.1 -U tester ${dbName}")
-            '';
```
srid, line 64: Oh, can we actually add the new native test to the example?

shivaraj-bh:
@srid Looks like the CI is running forever when I set `disabled = true;` for test process

srid (approved):
Let's get this merged, after switching upstream flake url.

shivaraj-bh:
Done!

PR #40, opened 2023-08-10
garnix ci: Enable macOS builds

srid:


srid:
Alas, we can't use garnix yet - https://github.com/garnix-io/issues/issues/61

PR #44, opened 2023-08-26
`mkPackageFor` in test flake

shivaraj-bh:


test/flake.nix
```diff
@@ -18,32 +18,30 @@
...
-          };
-        };
+        process-compose =
+          let
+            mkPackageFor = mod:
+              let
+                # Derive name from filename
+                name = lib.pipe mod [
+                  builtins.baseNameOf
+                  (builtins.match "(.*)_test.nix")
```
srid, line 28: Be careful with `builtins.match` regex parsing. IIRC it has issues on darwin, but not on Linux (going by what @ursi once told me).

srid (approved):

test/flake.nix
srid: I think it's fine if we have CI for both platforms already.

PR #57, opened 2023-09-29
Namespaces

### LINKED ISSUE #46: Assign a `namespace` to each service group?
PC has a `namespace` column:

image

Which appears to come from the `namespace` yaml option,
https://github.com/search?q=repo%3AF1bonacc1%2Fprocess-compose%20namespace&type=code

I think it makes sense to put each process under`services.foo."name1"` in its own namespace called "name1".

Support for this was added on Jun 10th: https://github.com/F1bonacc1/process-compose/releases/tag/v0.51.0 ... I wonder how namespaces work with other features, like depends_on.

shivaraj-bh:
Closes #46 

Question: Won't it be repetitive to have the process name and namespace present the same info? For instance in the screenshot below, `pg1` and `pg1-init` are grouped together under `pg1` namespace, which is the intended behaviour, but doesn't it look like there are too many `pg1's`.
Screenshot 2023-09-29 at 3 28 18 PM

TODO:
- [x] Change process-compose-flake input to upstream post this [PR](https://github.com/Platonic-Systems/process-compose-flake/pull/45).
- Use [this](https://github.com/F1bonacc1/process-compose/blob/5d7d23deca386ea456d76effb594a7c326643375/README.md?plain=1#L683-L686) to create a small clip and demonstrate how to start selective namespaces, when others aren't needed.

### DIFF
--- a/example/flake.lock
+++ b/example/flake.lock
@@ -70,15 +70,16 @@
     },
     "process-compose-flake": {
       "locked": {
-        "lastModified": 1693927910,
-        "narHash": "sha256-qPKHnWWzHS2bAi/SsFePQkGFeC2E1jklUjEidfQwYLc=",
-        "owner": "Platonic-Systems",
+        "lastModified": 1695980677,
+        "narHash": "sha256-5tHNbk0ldLUjAqKRZog/3asiVvkD51VGK9TvwzUBs38=",
+        "owner": "shivaraj-bh",
         "repo": "process-compose-flake",
-        "rev": "5494afa0b6a7bc4ccf82ef1c36fe1fcdb4217255",
+        "rev": "cad5d70f52727ed0a828b69a7e9949f925585b99",
         "type": "github"
       },
       "original": {
-        "owner": "Platonic-Systems",
+        "owner": "shivaraj-bh",
+        "ref": "namespace",
         "repo": "process-compose-flake",
         "type": "github"
       }
--- a/example/flake.lock
+++ b/example/flake.lock
@@ -70,16 +70,15 @@
     },
     "process-compose-flake": {
       "locked": {
-        "lastModified": 1695980677,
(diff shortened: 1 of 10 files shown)


srid (commented):
Regarding that question ... now that we have namespaces (derived from service name), can we stop using the service name prefix in process names? For eg., instead of `pg1`, `pg1-init` we would have the constant names `postgres` and `postgres-init`. Are there any downsides to doing it? These constant names are also more descriptive of what the process is actually.

shivaraj-bh:
> Regarding that question ... now that we have namespaces (derived from service name), can we stop using the service name prefix in process names? For eg., instead of `pg1`, `pg1-init` we would have the constant names `postgres` and `postgres-init`. Are there any downsides to doing it? These constant names are also more descriptive of what the process is actually.

The problem with this is, even though process-compose might support having processes with same name but different namespace, process-compose-flake doesn't (as type of processes is an attrset); should we switch to making processes a list instead?

srid:
Oh I had overlooked that.

> even though process-compose might support having processes with same name

Does it. Processes are specified as a map in [its YAML config](https://github.com/F1bonacc1/process-compose#quick-start), so I think not? If so, ignore my suggestion.

Anyway, back to your original question I wouldn't worry repetitiveness just yet.

shivaraj-bh:
> Does it? Processes are specified as a map in [its YAML config](https://github.com/F1bonacc1/process-compose#quick-start), so I think not? If so, ignore my suggestion.

That's right. I just ran a quick test and observed that, if there are multiple processes with same name it just chooses the process that is mentioned last in the yaml configuration.

Is it worth creating an issue in the upstream to solve this? ideally, different namespaces should be allowed to have processes with same name.

shivaraj-bh:
Good to merge?

srid (approved):

Issue #63, opened 2023-11-28
Postgres: use `psql` instead of `postgres` CLI to load the `initialScript`


shivaraj-bh:
SQL dumps can have multi-line queries, for instance:
```sql
CREATE TABLE users (
  id SERIAL PRIMARY KEY,
  name VARCHAR(50) NOT NULL,
  email VARCHAR(50) NOT NULL UNIQUE
);
```
In such a case using `postgres -E` as [here](https://github.com/juspay/services-flake/blob/48cd3d06c918b8104f47a5bfa27519652b93ecbf/nix/postgres.nix#L320-L322) will error out because the command expects single line statements. 

The solution to which can be `postgres -j -E` but I think it is ideal to load the dumps with `psql` command instead as suggested in the [postgres official doc](https://github.com/juspay/services-flake/blob/48cd3d06c918b8104f47a5bfa27519652b93ecbf/nix/postgres.nix#L320-L322).

## Proposed solution
**Context:** The reason for using `postgres` is that we won't have the postgres server running while the init script is running. 

We can run `initialScript.before` using `postgres` (as init script, before starting the server) and have `initialScript.after` run with `psql` as a process dependent on the postgres server. 

Ideally, to keep things uniform, I would prefer running any init SQL commands via psql.

Issue #64, opened 2023-12-13
Postgres: accept list of init `.sql` dumps


shivaraj-bh:
Probably like this:
```nix
services.postgres.pg1 = {
  enable = true;
  initDumps = [
    ./dump1.sql
    ./dump2.sql
  ];
};
```

PR #66, opened 2023-12-20
Using psql command instead of postgres in initialScript

rsrohitsingh682:
Resolves : https://github.com/juspay/services-flake/issues/63

Key changes in this PR:
- Use `psql` instead of `postgres`, as a result use `pg_ctl` to temporarily start the postgres server in `${name}-init` process.
- Avoid creating a custom `postgresPkg` with extensions by using [apply](https://github.com/juspay/services-flake/pull/66/commits/6940b2ca4986ae73c29d7f69f25fa3b4e5b894c2) attribute provided by `mkOption`
- Refactor setup script, use `writeShellApplication` and resolve shell-check errors


shivaraj-bh:
The CI is going into an infinite loop while running the `postgres_test.nix`, stopping it for now.

example/flake.nix
```diff
@@ -51,7 +51,7 @@
               {
                 environment.PGWEB_DATABASE_URL = "postgres://$USER@${pgcfg.listen_addresses}:${builtins.toString pgcfg.port}/${dbName}";
                 command = pkgs.pgweb;
-                depends_on."pg1".condition = "process_healthy";
+                depends_on."pg1-init".condition = "process_completed_successfully";
```
srid, line 54: Why this change?

shivaraj-bh: Earlier we would use `postgres` CLI command to load the `initialScript` but as explained in https://github.com/juspay/services-flake/issues/63, it might not be the ideal way to load dumps/scripts. Switching to `psql` to do the same would require the `pg1-init` to run after `pg1`, hence the dependency changes. Although the naming of `pg1-init` should be changed as it is misleading.

shivaraj-bh: This would prove to be useful while resolving https://github.com/juspay/services-flake/issues/64 as `psql` works well with most of the dumps, but the `postgres` CLI command doesn't as seen in our case while trying to load these dumps: https://github.com/nammayatri/nammayatri/blob/94d8212c20043ad179e5b51b599deab4af8a4e8b/Backend/nix/arion-configuration.nix#L23-L38

shivaraj-bh: @srid I have added comments to move back to depending on `pg1` as we can use `pg_ctl` to temporarily start the postgres server while `pg1-init` is running.

shivaraj-bh: I have also refactored the service a bit, a lot of it can still be split into parts but this seemed enough based on what the PR is titled.

srid: LGTM

shivaraj-bh (approved):

shivaraj-bh:
@rsrohitsingh682 will merge this today if there is nothing more to add

PR #67, opened 2023-12-22
Adding option initialDumps to Postgres

rsrohitsingh682:
Added an option initialDumps which can be used to load lists of SQL dumps during the database initialization.


shivaraj-bh:
This will only work after #66

shivaraj-bh:
@rsrohitsingh682 You will have to rebase with main, some structural changes were done to the postgres service in #66

nix/postgres/default.nix
```diff
@@ -229,6 +229,15 @@ in
         ]
       '';
     };
+    
+    initialDumps = lib.mkOption {
+      type = types.listOf types.path;
+      default = [ ];
+      description = "List of SQL dumps to run during the database initialization.";
```
shivaraj-bh, line 236:
```suggestion
      description = ''List of SQL dumps to run during the database initialization.
      These dumps are loaded after `initalScript` and `initialDatabases`, if there is any conflict, it is advised to move the queries from your `.sql` files to one of these options.'';
```

shivaraj-bh (changes requested):
Let’s include a test for `initialDatabases`

shivaraj-bh (approved):

Issue #70, opened 2023-12-26
Postgres: add `depends_on` option


shivaraj-bh:
This option will be useful to append extra `depends_on` configuration to the `<name>-init` process. 

An example scenario is when a replica DB has to depend on another process that runs `pg_basebackup`.

srid:
Can I see the place where this gets actually _used_?

PR #71, opened 2023-12-26
Postgres: Added option depends_on

rsrohitsingh682:
Resolves: #70

### DIFF
--- a/nix/postgres/default.nix
+++ b/nix/postgres/default.nix
@@ -230,6 +230,24 @@ in
       '';
     };
 
+    depends_on = lib.mkOption {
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
@srid This option is a copied from [process-compose-flake](https://github.com/Platonic-Systems/process-compose-flake/blob/9e77499a9f4572d91771b21da17006f24950185c/nix/process-compose/settings/process.nix#L22-L38), does this look good or is there a cleaner way to reuse the option?

srid (commented):
Copying is fine, but you want to flesh out the `description` a bit to be meaningful and specific.

PR #73, opened 2024-01-13
[Postgres]: Use `schemas` in-place of `initialDumps`

shivaraj-bh:
## Problem

`initialDumps` is always created under `postgres` database, which is not a desirable default behaviour when your dumps are intended for different database, like [`atlas_dev` in nammayatri](https://github.com/nammayatri/nammayatri/blob/2e20e20b89591501b32743c4cfe52b24dd1aabda/Backend/nix/arion-configuration.nix#L49).

## Solution

We already have [schema](https://github.com/juspay/services-flake/blob/2211042f3e04de6b105f8d70c0be6213a2127d88/nix/postgres/default.nix#L207-L214) module option to load a schema into a given [database name](https://github.com/juspay/services-flake/blob/2211042f3e04de6b105f8d70c0be6213a2127d88/nix/postgres/default.nix#L201-L206). The schema module option already supports loading from either a file or all the [`.sql` files in a folder.](https://github.com/juspay/services-flake/blob/2211042f3e04de6b105f8d70c0be6213a2127d88/nix/postgres/setup-script.nix#L31-L33) 

The sql dumps in some cases can be [scattered across folders](https://github.com/nammayatri/nammayatri/blob/2e20e20b89591501b32743c4cfe52b24dd1aabda/Backend/nix/arion-configuration.nix#L23-L38), and hence we can rename `schema` to `schemas` and allow it. 

Tagging @rsrohitsingh682 here as he had worked on [`initialDumps`](https://github.com/juspay/services-flake/pull/67)

### DIFF
--- a/nix/postgres/postgres_test.nix
+++ b/nix/postgres/postgres_test.nix
@@ -34,7 +34,7 @@
           # initialScript.after test
           echo "SELECT 1 FROM pg_database WHERE datname = 'foo';" | psql -h 127.0.0.1 | grep -q 1
 
-          #intialDumps test
+          # schemas test
           echo "SELECT * from users where user_name = 'test_user';" | psql -h 127.0.0.1 -p 5433 -d sample-db | grep -q test_user
         '';
         name = "postgres-test";
--- a/nix/postgres/postgres_test.nix
+++ b/nix/postgres/postgres_test.nix
@@ -4,7 +4,18 @@
     listen_addresses = "127.0.0.1";
     initialScript.before = "CREATE USER bar;";
     initialScript.after = "CREATE DATABASE foo OWNER bar;";
-    initialDumps = [ ./test.sql ];
+  };
+  services.postgres."pg2" = {
+    enable = true;
+    port = 5433;
+    listen_addresses = "127.0.0.1";
+    # INFO: pg1 creates $USER database while pg2 doesn't because `initialDatabases` is present
+    initialDatabases = [
+      {
+        name = "sample-db";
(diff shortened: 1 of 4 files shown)


nix/postgres/postgres_test.nix
```diff
@@ -24,7 +35,7 @@
           echo "SELECT 1 FROM pg_database WHERE datname = 'foo';" | psql -h 127.0.0.1 | grep -q 1
 
           #intialDumps test
```
srid, line 37: out of date comment

nix/postgres/setup-script.nix
```diff
@@ -1,5 +1,27 @@
 { config, pkgs, lib }:
 let
+  setupInitialSchema = dbName: schema: ''
+    ${lib.optionalString (schema != null) ''
+    echo "Applying database schema on ${dbName}"
+    if [ -f "${schema}" ]
+    then
+      echo "Running file ${schema}"
+      awk 'NF' "${schema}" | psql -d ${dbName}
```
srid, line 9: This uses `/usr/bin/awk` on macOS? I'd think `lib.getExe pkgs.gawk` would be more reproducible.

srid (approved):

nix/postgres/setup-script.nix
shivaraj-bh: It is actually run inside a script that gets its runtime inputs from [here](https://github.com/juspay/services-flake/blob/2211042f3e04de6b105f8d70c0be6213a2127d88/nix/postgres/setup-script.nix#L86)

Issue #74, opened 2024-01-13
CI hangs infinitely


shivaraj-bh:
It seems to happen while running `default-test` from [example/flake.nix](https://github.com/juspay/services-flake/blob/main/example/flake.nix). This isn’t something that happens always but whenever it does, I will have to manually restart the workflow. 

See an example workflow here: https://github.com/juspay/services-flake/actions/runs/7513430525/job/20455173179?pr=73

I can’t seem to reproduce this in my local atm.

shivaraj-bh:
Maybe not only at `default-test`: https://github.com/juspay/services-flake/actions/runs/7513827171/job/20456036196

Either something is wrong with github actions or process-compose test process is failing to close after completion.

shivaraj-bh:
I am suspecting it could most likely be due to port collisions while running checks parallely, as this is the only thing that is shared across processes. In this case, we might have to do something that [nixosTests](https://github.com/NixOS/nixpkgs/blob/e3528ac5517c68ef67bce66b9dce5433a402ceda/nixos/tests/kafka.nix#L80) does.

If this is the case, I should be able to reproduce locally by trying `nix flake check` multiple times but without using the `/nix/store` cache, is there a way?

srid:
`--rebuild`?

shivaraj-bh:
`nix flake check` doesn’t support `—rebuild`

shivaraj-bh:
I might have stumbled upon a fix for this issue while trying to flesh out some documentation for `services-flake`. Turns out if you follow the instructions in [`Getting Started` section](https://github.com/juspay/services-flake/blob/c7bed59b874f367813966725ae390691c65cedff/README.md?plain=1#L13-L17), on the first run `pg1` is skipped because `pg1-init` fails to complete with an `initdb` error complaining about “failing to create `data` directory as it already exists". Then I scrolled up the logs in CI for a few builds that were running infinitely and [found the same error](https://github.com/juspay/services-flake/actions/runs/7558073032/job/20578768690?pr=77#step:6:4346).

Of course the `test` process depends on both `pg1` and `pg2` being up and running, and since that is not the case above, it hangs infinitely. 

**Why does it not happen all the time?**
It all depends on how the processes are spawned, it happens only when both `pg2-init` and `pg1-init` tries to create `data` directory at the same time.

shivaraj-bh:
[...]

PR #77, opened 2024-01-17
postgres: add socketDir option

johnalotoski:
* Postgres: Fix a build failure when initialDatabase schema is null
* Postgres: Add a socketDir config option which defaults to config.dataDir and allows customization of socketDir for users who do not wish it fixed to dataDir
  * Also, sockets are limited to around [100 chars](https://linux.die.net/man/7/unix)
  * In the case that the default or desired dataDir absolute path exceeds this char limit, postgres will fail to start if it is also used for socketDir
  * In this case, dataDir can remain as desired and socketDir can be set separately to a shorter absolute path

### DIFF
--- a/nix/postgres/setup-script.nix
+++ b/nix/postgres/setup-script.nix
@@ -1,7 +1,6 @@
 { config, pkgs, lib }:
 let
   setupInitialSchema = dbName: schema: ''
-    ${lib.optionalString (schema != null) ''
     echo "Applying database schema on ${dbName}"
     if [ -f "${schema}" ]
     then
@@ -20,7 +19,6 @@ let
       echo "ERROR: Could not determine how to apply schema with ${schema}"
       exit 1
     fi
-    ''}
   '';
   setupInitialDatabases =
     if config.initialDatabases != [ ] then
@@ -37,7 +35,8 @@ let
           if [ 1 -ne "$dbAlreadyExists" ]; then
             echo "Creating database: ${database.name}"
             echo 'create database "${database.name}";' | psql -d postgres
-            ${lib.concatMapStrings (schema: setupInitialSchema (database.name) schema) database.schemas}
+            ${lib.optionalString (database.schemas != null)
+              (lib.concatMapStrings (schema: setupInitialSchema (database.name) schema) database.schemas)}
           fi
         '')
         config.initialDatabases)
--- a/nix/postgres/setup-script.nix
+++ b/nix/postgres/setup-script.nix
@@ -100,7 +100,7 @@ in
       echo
       echo "PostgreSQL is setting up the initial database."
       echo
-      PGHOST=$(mktemp -d "$(readlink -f ${config.dataDir})/pg-init-XXXXXX")
+      PGHOST=$(mktemp -d "$(readlink -f "${config.socketDir}")/pg-init-XXXXXX")
       export PGHOST
 
       function remove_tmp_pg_init_sock_dir() {
--- a/nix/postgres/default.nix
+++ b/nix/postgres/default.nix
@@ -50,6 +50,12 @@ in
       description = "The DB data directory";
     };
 
+    socketDir = lib.mkOption {
+      type = lib.types.str;
+      default = config.dataDir;
+      description = "The DB socket directory";
+    };
+
     hbaConf =
       let
         hbaConfSubmodule = lib.types.submodule {
@@ -161,7 +167,7 @@ in
         default = {
           listen_addresses = config.listen_addresses;
           port = config.port;
-          unix_socket_directories = config.dataDir;
+          unix_socket_directories = config.socketDir;
           hba_file = "${config.hbaConfFile}";
         };
       };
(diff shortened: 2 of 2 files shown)


nix/postgres/setup-script.nix
```diff
@@ -101,7 +100,7 @@ in
       echo
       echo "PostgreSQL is setting up the initial database."
       echo
-      PGHOST=$(mktemp -d "$(readlink -f ${config.dataDir})/pg-init-XXXXXX")
+      PGHOST=$(mkdir -p "${config.socketDir}" && mktemp -d "$(readlink -f "${config.socketDir}")/pg-init-XXXXXX")
```
shivaraj-bh, line 103: Why do we have to `mkdir` here? wouldn’t just `mktemp` suffice? because we are using this only to temporarily start the server during initialisation and we delete the socket after init.

johnalotoski: You are correct; this was left over from a previous command modification and is no longer needed.
Now to try and figure out why the CI test seems to hang and fix that as well.

johnalotoski: The CI test that failed earlier fails on what looks like a race condition for the postgres test when both `./data/pg1` and `./data/pg2` are created at the same time, and one of the two fails to start because of a "File exists" error.  Ie:
```
default-test> [pg2-init ] initdb: error: could not create directory "./data": File exists
default-test> [pg1-init ] creating directory ./data/pg1 ... ok
default-test> [pg2-init ] creating directory ./data/pg2 ...
...
<pg2 then fails to start in this example>
```

I can reproduce the failure locally with nixci, but I cannot reproduce it by executing the commands manually outside of the nixci test script, or by loading the same test setup into a process-compose yaml and running it.

We could change the postgresql_test.nix file to use two separate parent dataDirs which resolves this race condition test failure, like:
```
  services.postgres."pg1" = {
    enable = true;
+   dataDir = "./data1/pg1";
    ...
  };
  services.postgres."pg2" = {
    enable = true;
+   dataDir = "./data2/pg2";
    ...
  };
```

When I test with nixci locally using the main services-flake branch, it also fails this way.  Strangely, the CI history doesn't seem to show it failing like this in the past.  Have you encountered this before?

shivaraj-bh: Actually, yes. This is the reason for https://github.com/juspay/services-flake/issues/74


I have also raised a fix for this: https://github.com/juspay/services-flake/pull/78

Now the CI is failing due to an unrelated error: “Error: devour-flake produced an outpath (/nix/store/1s02vc584sy7nzhx81j93mk1mrwnql9g-devour-output) with no outputs”

shivaraj-bh: And this seems to be happening consistently across all the CI runs. I can’t reproduce it locally either.

srid: > Now the CI is failing due to an unrelated error: “Error: devour-flake produced an outpath (/nix/store/1s02vc584sy7nzhx81j93mk1mrwnql9g-devour-output) with no outputs”

Looks like Nix 2.19 had a regression. See https://github.com/srid/devour-flake/issues/22

Is there an official bug report? Or was there a backwards incompatible change?

srid: Should be fixed (in our repo) after https://github.com/juspay/services-flake/pull/79

shivaraj-bh:
@johnalotoski You can rebase with main and the CI should pass

johnalotoski:
Rebased and ready for workflow approval.

shivaraj-bh:
LGTM! Merging

nix/postgres/setup-script.nix
johnalotoski: @shivaraj-bh, apologies for revisiting this, but back to the original question of why do we have mkdir here:
```
-     PGHOST=$(mktemp -d "$(readlink -f ${config.dataDir})/pg-init-XXXXXX")
+     PGHOST=$(mkdir -p "${config.socketDir}" && mktemp -d "$(readlink -f "${config.socketDir}")/pg-init-XXXXXX")
```

When configDir and socketDir are the same (the default), the `initDb -D ${config.dataDir}` earlier in the postgres setup-script does the equivalent of a `mkdir -p` for config.dataDir which is why mktemp in that case doesn't require it.

It's also not a problem when configDir and socketDir are different and the full path for the socketDir already exists, no mkdir -p is required.

In the remaining case where configDir and socketDir are different and socketDir path doesn't exist yet or only exists partially, mktemp only tries to create the leaf node directory, but not any other part of the parental dir path and so it fails.
```
❯ mktemp -d 1/2/3/pg-init-XXXXXX
mktemp: failed to create directory via template ‘1/2/3/pg-init-XXXXXX’: No such file or directory

❯ mkdir -p 1/2/3 && mktemp -d 1/2/3/pg-init-XXXXXX
1/2/3/pg-init-Vwl7c7
```

When I commented yesterday it wasn't needed anymore, I was no longer testing this 3rd case and didn't realize it was still an issue, but I ran into it again today.  I'll PR a line line update shortly.

shivaraj-bh: Makes sense!

PR #81, opened 2024-01-18
postgres ensure socketDir mktemp init succeeds

johnalotoski:
* Ensure postgres socketDir mktemp initialization succeeds when the socketDir path does not exist yet
  * See https://github.com/juspay/services-flake/pull/77#discussion_r1458072831
* Add a test for socketDir configured with a multi-level non-existing path

### DIFF
--- a/nix/postgres/postgres_test.nix
+++ b/nix/postgres/postgres_test.nix
@@ -7,6 +7,7 @@
   };
   services.postgres."pg2" = {
     enable = true;
+    socketDir = "./test/new/socket/path";
     port = 5433;
     listen_addresses = "127.0.0.1";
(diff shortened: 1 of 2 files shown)


shivaraj-bh:
Can we also add a test for the scenario where `socketDir` is different from `dataDir`? `pg2` in `nix/postgres/postgres_test` can be used for this maybe? even `pg1` works. Should’ve done this in the last PR perhaps, would’ve been caught in CI.

johnalotoski:
Good idea, I can look at doing that tomorrow.

johnalotoski:
Added the test @shivaraj-bh.

shivaraj-bh:
Great! 

Just to confirm I ran the test with the current main, and as expected it does fail:
Screenshot 2024-01-20 at 12 18 08 AM

Looks good

PR #87, opened 2024-01-24
docs/clickhouse: add intro

srid:
Emanote uses the first paragraph as page description, thus it is useful to write that paragraph reflecting the description of that page. Without this PR, link previews look like this:

image

^ which is not the correct description for the page.


srid:
FTR, the para is from https://en.wikipedia.org/wiki/ClickHouse

PR #91, opened 2024-02-06
clickhouse init script

roman-bodavskiy:


shivaraj-bh:
can you also include a test for `initialDatabases` here: https://github.com/juspay/services-flake/blob/main/nix/clickhouse_test.nix

roman-bodavskiy:
> can you also include a test for `initialDatabases` here: https://github.com/juspay/services-flake/blob/main/nix/clickhouse_test.nix

resolved

nix/clickhouse/default.nix
```diff
@@ -0,0 +1,165 @@
...
+                  name = "setup-clickhouse";
+                  runtimeInputs = with pkgs; [ config.package coreutils gnugrep gawk ];
+                  text = ''
+                    if test -d ${config.dataDir}
+                      then echo "Clickhouse database directory ${config.dataDir} appears to contain a database; Skipping initialization"
+                      else
+                        echo "Clickhouse is setting up the initial database."
+                        set -m
+                        clickhouse-server --config-file=${clickhouseConfig} &
+                        sleep 5s
```
shivaraj-bh, line 119: Starting `clickhouse-server` in background with `&` and waiting for 5s, which we aren’t even sure if its going to be enough, isn’t a good way to initialise. Although I tried a couple of things:
- `clickhouse-server —daemon`: doesn’t seem to work at all and I have no clue why.
- `clickhouse start` should start in background according to [docs](https://clickhouse.com/docs/en/install#launch) but it expects sudo as it tries to create `/var/run/clickhouse-server` and I don’t see any way to override this.

As I don’t see any better approach for now, we can keep this implementation as is and visit this in future.

roman-bodavskiy: > Starting clickhouse-server in background with & and waiting for 5s, which we aren’t even sure if its going to be enough, isn’t a good way to initialise.
anyway if 5 secs will be not enough, then init script will fail and main process will not start, and we will face error. I've tested several times, and all worked fine

shivaraj-bh: Yes, that’s true but some machines might be too slow to even initialise in 5 seconds or someone might use a configuration that might cause delay in initialisation. Anyway, we can look at it if that ever happens.

PR #94, opened 2024-02-11
Revert "ci: Don't upgrade Nix behind the scenes"

srid:
Reverts juspay/services-flake#79

To fix, https://github.com/srid/nixci/issues/35

Also, introduces Garnix cache to speed up fetching of nixci.


srid:
oof 

image

PR #103, opened 2024-02-19
explicitly define project root in treefmt

conscious-puppet:
specifying `projectRoot` in `treefmt` fixes #45 

```
treefmt-check> treefmt 0.6.1
treefmt-check> [INFO ] #nixpkgs-fmt: 26 files processed in 31.84ms
treefmt-check> 0 files changed in 37ms (found 42, matched 26, cache misses 26)
treefmt-check> On branch main
treefmt-check> nothing to commit, working tree clean
```

PR #104, opened 2024-02-19
add format check for services flake in dev/flake.nix

conscious-puppet:
adding `check` in `test/flake.nix` checks for the formatting across services flake

```
❯ nix run github:srid/nixci --option sandbox true .
```

```
treefmt-check> Initialized empty Git repository in /build/project/.git/
treefmt-check> Initialized empty Git repository in /build/project/.git/
treefmt-check> treefmt 0.6.0
treefmt-check> 0 files changed in 2ms (found 2, matched 1, cache misses 1)
treefmt-check> On branch main
treefmt-check> nothing to commit, working tree clean
treefmt-check> treefmt 0.6.0
treefmt-check> 0 files changed in 20ms (found 42, matched 26, cache misses 26)
treefmt-check> On branch main
treefmt-check> nothing to commit, working tree clean
```

### DIFF
--- a/test/flake.nix
+++ b/test/flake.nix
@@ -5,19 +5,28 @@
     systems.url = "github:nix-systems/default";
     process-compose-flake.url = "github:Platonic-Systems/process-compose-flake";
     services-flake.url = "github:juspay/services-flake";
+    treefmt-nix.url = "github:numtide/treefmt-nix";
   };
   outputs = inputs:
     inputs.flake-parts.lib.mkFlake { inherit inputs; } {
       systems = import inputs.systems;
       imports = [
         inputs.process-compose-flake.flakeModule
+        inputs.treefmt-nix.flakeModule
       ];
-      perSystem = { self', pkgs, system, lib, ... }: {
+      perSystem = { self', pkgs, system, lib, config, ... }: {
         _module.args.pkgs = import inputs.nixpkgs {
           inherit system;
           # Required for elastic search
           config.allowUnfree = true;
         };
+        treefmt = {
+          projectRootFile = "flake.nix";
+          programs = {
+            nixpkgs-fmt.enable = true;
+          };
+        };
+        checks.fmt-check = config.treefmt.build.check inputs.services-flake;
         process-compose =
           let
             mkPackageFor = mod:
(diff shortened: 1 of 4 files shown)


test/flake.nix
```diff
@@ -5,19 +5,28 @@
...
       ];
-      perSystem = { self', pkgs, system, lib, ... }: {
+      perSystem = { self', pkgs, system, lib, config, ... }: {
         _module.args.pkgs = import inputs.nixpkgs {
           inherit system;
           # Required for elastic search
           config.allowUnfree = true;
         };
+        treefmt = {
+          projectRootFile = "flake.nix";
```
shivaraj-bh, line 24:
```suggestion
          projectRoot = inputs.services-flake;
          projectRootFile = "flake.nix";
```

Like this you might not need to create another check

conscious-puppet: sure i'll give it a try

shivaraj-bh:
Looks like it is working fine (from ubuntu-latest CI job):
Screenshot 2024-02-20 at 12 56 24 AM

test/flake.nix
conscious-puppet: @shivaraj-bh this suggestion works. Although I had another concern. I am having to define `treefmt.programs` twice. once in `dev/flake.nix` and then again in `test/flake.nix`. Is there a way I could get the programs as well from the `${inputs.services-flake}/dev/flake.nix`

shivaraj-bh: Why not just add `services-flake` in inputs of `./dev`? and it should work without having to make any changes in `./test/flake.nix`

conscious-puppet: sure i'll try that only

shivaraj-bh: If this works out you can make the change in #103

conscious-puppet: adding to `services-flake` to `./dev/flake.nix` works. i've pushed that change

conscious-puppet: > If this works out you can make the change in https://github.com/juspay/services-flake/pull/103

i missed your comment before pushing to this branch. i'll close this one and push in #103 after the build is successful here

PR #108, opened 2024-02-21
init grafana service

### LINKED ISSUE #59: Add Grafana service

conscious-puppet:
resolves #59


nix/grafana/default.nix
```diff
@@ -0,0 +1,83 @@
...
+                grafana server --config ${grafanaConfigIni} \
+                               --homepath ${config.package}/share/grafana \
+                               cfg:paths.data="$(readlink -m ${config.dataDir})"
+              '';
+            };
+          in
+          {
+            command = startScript;
+            readiness_probe = {
+              exec.command = "${pkgs.curl}/bin/curl -f ${grafanaConfig.server.protocol}://${grafanaConfig.server.domain}:${builtins.toString grafanaConfig.server.http_port}/api/health";
```
shivaraj-bh, line 68: You can use the [http_get](https://github.com/Platonic-Systems/process-compose-flake/blob/b346fb30ad5fdee14e3edc3ba490eaa5db9ec0b7/nix/process-compose/settings/probe.nix#L13-L37) readiness probe instead of curl.

See example: https://github.com/Platonic-Systems/process-compose-flake/blob/b346fb30ad5fdee14e3edc3ba490eaa5db9ec0b7/example/flake.nix#L56-L59

nix/grafana/default.nix
```diff
@@ -0,0 +1,83 @@
...
+        processes."${name}" =
+          let
+            grafanaConfig =
+              {
+                server = {
+                  protocol = "http";
+                  http_port = config.port;
+                  domain = "localhost";
+                };
+              } // config.extraConf;
```
srid, line 53: What happens if `extraConfg = { server.http_port = 8080; }`? Would it not end up removing `protocol` and `domain`? Shouldn't we do a deep merge?

nix/grafana/default.nix
srid: Also, perhaps make the whole `${grafanaConfig.server.protocol}://${grafanaConfig.server.domain}:${builtins.toString grafanaConfig.server.http_port}"` a value of an internal option (`grafanaUri`?), so it can be re-used in other places, including by the user. It also makes the probe CLI here more readable.

conscious-puppet: @srid sure.

grafana has a config called `root_url` for this. i'll store the resulting value in this config

nix/grafana/default.nix
conscious-puppet: i am using `lib.recursiveUpdate` now.

i have made this same error here. i'll create a PR to fix this

- https://github.com/juspay/services-flake/blob/76beca31cc5c257f02b0fbb4433803ab4bab6a32/nix/clickhouse/default.nix#L95

nix/grafana/default.nix
```diff
@@ -0,0 +1,105 @@
...
+      type = types.str;
+      description = "Protocol (http, https, h2, socket).";
+      default = "http";
+    };
+
+    root_url = lib.mkOption {
+      type = types.str;
+      description = "The full public facing url.";
+      default = "${config.protocol}://${config.domain}:${builtins.toString config.http_port}";
+    };
```
srid, line 37: If we don't expect the user to set this, you want to add `readOnly = true;` here.

conscious-puppet: grafana actually allows to give a `root_url` with subpath. in that case, user may set this value.

ref: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#root_url

shivaraj-bh: Will the user simply not set `extraConfig.server.root_url` instead? This additional option might not be needed. I think the [latest impl. of readiness_probe](https://github.com/juspay/services-flake/pull/108/commits/98811170929b85415a9e2c42b0e03a0f402435fd) looks good as is

doc/grafana.md
```diff
@@ -0,0 +1,62 @@
...
+    };
+  };
+};
+```
+
+3. Add a setting to start `grafana` only after `postgres` is running.
+
+```nix
+settings.processes."gf1".depends_on."pg1".condition = "process_healthy";
+```
```
shivaraj-bh, line 62:
````suggestion
```nix
{
  services.postgres.pg1 = {
    enable = true;
    listen_addresses = “127.0.0.1”;
  };
  services.grafana.gf1 = {
    enable = true;
    extraConf.database = with config.services.postgres.pg1; {
      type = “postgres”;
      host = “${listen_addresses}:${port}”;
    };
  };
  
  settings.processes."gf1".depends_on."pg1".condition = "process_healthy";
}
```
````

It would be nice to keep these tips/suggestions simple and short. 
I have assumed that not specifying name and user for postgres db in grafana will use postgres by default for both, if not we can explicitly mention.

Comment (shivaraj-bh) [reply doc/grafana.md:62]: `initialDatabases` and `intialScript` can be part of postgres doc (whenever we add it) and if one wants to play around with postgres configs, they can refer from there.

I say this because, we are better off keeping these mini-tutorials short (with default configs) and not overwhelm the reader.

Comment (conscious-puppet) [reply doc/grafana.md]: sure, by default grafana looks for the user `root` in database. and for the database `grafana`. i have added config these two over your suggestion. please have a look

File: doc/grafana.md [hunk 5]
```diff
@@ -0,0 +1,40 @@
+# Grafana Open Source
```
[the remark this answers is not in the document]
conscious-puppet: i copied the doc from here:

- https://grafana.com/docs/grafana/latest/fundamentals/#grafana-open-source

i think, they probably want to distinguish this from `Grafana Cloud`. I'll it to grafana, sure

PR #110, opened 2024-02-22
clickhouse: use `lib.recursiveUpdate` to deep merge extraConfig

conscious-puppet:
previously, was shallow merge to append extra configs, which would have removed some configs. updated to use `lib.recursiveUpdate` to deep merge extraConfig.

### DIFF
--- a/nix/clickhouse/default.nix
+++ b/nix/clickhouse/default.nix
@@ -27,7 +27,7 @@ in
       description = "The clickhouse data directory";
     };
 
-    defaultConfig = lib.mkOption {
+    defaultExtraConfig = lib.mkOption {
       type = yamlFormat.type;
       internal = true;
       readOnly = true;
@@ -97,7 +97,7 @@ in
         processes =
           let
             clickhouseConfig = yamlFormat.generate "clickhouse-config.yaml" (
-              lib.recursiveUpdate config.defaultConfig config.extraConfig
+              lib.recursiveUpdate config.defaultExtraConfig config.extraConfig
             );
           in
           {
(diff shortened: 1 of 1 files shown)


nix/clickhouse/default.nix
```diff
@@ -75,24 +75,26 @@ in
...
+                  path = "${config.dataDir}/clickhouse";
+                  tmp_path = "${config.dataDir}/clickhouse/tmp";
+                  user_files_path = "${config.dataDir}/clickhouse/user_files";
+                  format_schema_path = "${config.dataDir}/clickhouse/format_schemas";
+                  user_directories = {
+                    users_xml = {
+                      path = "${config.package}/etc/clickhouse-server/users.xml";
+                    };
+                  };
+                }
```
srid, line 97: An `internal` and `readOnly` option containing this default value would improve a bit on readability.

conscious-puppet: sure, i'll add `internal` and `readOnly` options for the default config

conscious-puppet: updated with these changes

srid: Hmm, sorry I wasn't being clear. I meant *one* option like:

```nix
defaultExtraConfig = lib.mkOption {
  type = types.attrs;
  internal = true;
  readOnly = true;
  default = {
    tcp_port = "${toString config.port}";
    ...
  };
};
```

srid (approved):
`defaultExtraConfig` would be more clearer, other than that looks goods to me.

PR #111, opened 2024-02-22
Update process-compose-flake

srid:
To try the new JSON config, avoiding IFD.

### DIFF
--- a/test/flake.lock
+++ b/test/flake.lock
@@ -54,11 +54,11 @@
     },
     "process-compose-flake": {
       "locked": {
-        "lastModified": 1695992918,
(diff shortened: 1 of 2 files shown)


test/flake.lock
```diff
@@ -72,25 +72,9 @@
...
-        "repo": "services-flake",
-        "rev": "f171932d130c3ce5184b69144012a225b2e28868",
-        "type": "github"
-      },
-      "original": {
-        "owner": "juspay",
-        "repo": "services-flake",
-        "type": "github"
-      }
-    },
```
srid, line 93: A stale entry that had to be removed, due to use of `services-flake = {};`.

PR #116, opened 2024-02-29
cassandra: init service

conscious-puppet:
add cassandra service.


shivaraj-bh:
Can you also include the doc with a `Getting started` section, if there isn’t anything additional to add?

nix/cassandra.nix
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    listenAddress = lib.mkOption {
+      type = types.str;
+      description = "Listen address";
+      default = "127.0.0.1";
+      example = "127.0.0.1";
+    };
+
+    nativeTransportPort = lib.mkOption {
```
shivaraj-bh, line 26: Perhaps the `Getting started` can also mention how to change the port because it is not straightforward.

shivaraj-bh (approved):

PR #115, opened 2024-02-28
add pgAdmin service

conscious-puppet:
resolves #80 

**Related issues**

- https://github.com/NixOS/nixpkgs/issues/281214

PR #117, opened 2024-03-02
Update flake lock

conscious-puppet:


test/flake.nix
```diff
@@ -21,6 +22,8 @@
             (final: prev: {
               # Because tests are failing on darwin: https://github.com/juspay/services-flake/pull/115#issuecomment-1970467684
               pgadmin4 = prev.pgadmin4.overrideAttrs (_: { doInstallCheck = false; });
+              # grafana builds are failing on aarch64-darwin latest nixpkgs-unstable: https://github.com/NixOS/nixpkgs/issues/273998
+              grafana = inputs.nixpkgs-2023.legacyPackages.${system}.grafana;
```
conscious-puppet, line 26: @shivaraj-bh , i have pinned grafana to an older commit since it fails to build after updating `flake.lock`.

is this a good idea?

nix/pgadmin.nix
```diff
@@ -118,7 +118,7 @@ in
                       ${config.package}/bin/.pgadmin4-setup-wrapped setup
                     else
                       # pgadmin-8.2 has .pgadmin4-cli-wrapped
-                      ${config.package}/bin/.pgadmin4-cli-wrapped setup
+                      ${config.package}/bin/.pgadmin4-cli-wrapped setup-db
```
srid, line 121: What's the context for this change? Is services-flake not supported on nixpkgs older to this update anymore?

test/flake.nix
shivaraj-bh: To avoid confusion, we usually keep only one nixpkgs input. I think we can overlay grafana, in `nixpkgs` to a version that works. Let me try it out.

shivaraj-bh:
Since https://github.com/NixOS/nixpkgs/pull/283330, along with the go backend, grafana frontend is also built from source in nixpkgs. But the offline cache doesn’t support darwin, this can be fixed by overriding it:

```nix
{
  grafana = prev.grafana.overrideAttrs (oa: rec {
    offlineCache = oa.offlineCache.overrideAttrs (_ :{
      buildPhase = ''
        runHook preBuild
        export HOME="$(mktemp -d)"
        yarn config set enableTelemetry 0
        yarn config set cacheFolder $out
        yarn config set --json supportedArchitectures.os '[ “linux” “darwin" ]'
        yarn config set --json supportedArchitectures.cpu '[ “x64” “arm64" ]'
        yarn
        runHook postBuild
      '';
      outputHash = lib.fakeHash;
    });
  });
}
```

**Note:** replace the `outputHash` with whatever the offlinecache generates.

shivaraj-bh:
Even after fixing the frontend cache, the go backend fails to build due to: https://github.com/NixOS/nixpkgs/issues/273998

Looks like it is do with nix: https://github.com/NixOS/nix/issues/9977

shivaraj-bh:
As a workaround, we can use grafana < 10.2.0 until the case-insensitivity issue is resolved.

nix/pgadmin.nix
conscious-puppet: in the PR #115, i made a mistake with the setup command for pgadmin 8.2. since the older nixpkgs had pgadmin 7.5, i could not test this and it got missed.

for older nixpkgs, pgadmin has `pgadmin4-setup` executable to setup the configuration db, and for newer nixpkgs, it has `pgadmin4-cli` executable to setup the db. so services-flake would support both

```sh
  if [ -f ${config.package}/bin/.pgadmin4-setup-wrapped ]; then
    # pgadmin-7.5 has .pgadmin4-setup-wrapped
    ${config.package}/bin/.pgadmin4-setup-wrapped setup
  else
    # pgadmin-8.2 has .pgadmin4-cli-wrapped
    ${config.package}/bin/.pgadmin4-cli-wrapped setup-db
  fi
```

shivaraj-bh:
Turns out we can get the latest grafana build on darwin, thanks to: https://github.com/NixOS/nixpkgs/issues/129730

shivaraj-bh:
CI fails with: `no space left on device` 🙂

shivaraj-bh:
Until we self-host the runner, hope this helps: https://github.com/actions/runner-images/issues/2875

srid:
@conscious-puppet @shivaraj-bh What's the original reason for updating flake.lock in this PR? If we revert this, can we avoid having to build grafana on CI and locally?

conscious-puppet:
some of the packages were old.
- the default package for cassandra was cassandra 3. which had a dependency on python 2, and was no longer supported on aarch64-darwin
- pgadmin4 also had an outdated version with different cli for setup db.

if we revert this, then we may not be able to test services flake against current package versions with `just run <service>` or `just test <service>`.

conscious-puppet:
if build is an issue, can we overlay grafana to an older version for darwin?

srid:
> can we overlay grafana to an older version for darwin?

This is what I'm trying now

PR #121, opened 2024-03-06
Add pre-commit-hooks.nix, enforcing conventional commits

shivaraj-bh:
with treefmt and commitizen enabled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /result
 # created when `just run <service>` is used
 /test/data
+/.pre-commit-config.yaml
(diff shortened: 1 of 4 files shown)


srid (approved):
LGTM.  Are we Okay with the default config?

I guess we can tweak the rules going forward.

srid:
These commits can be squashed into a single one, since this PR does one thing, basically.

srid:
At some point, we should document our commit guidelines in https://community.flake.parts/services-flake/contributing

Issue #123, opened 2024-03-06
Link to the blog post


srid:
The docs should link to https://nixos.asia/en/blog/replacing-docker-compose in an appropriate section.

PR #125, opened 2024-03-07
docs: add motivation for creating services-flake

### LINKED ISSUE #124: docs: Add motivation / use case in README
https://twitter.com/4ngerman/status/1765496606005268675

Related: #123

shivaraj-bh:
resolves #124 and #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,14 @@ Declarative, composable, and reproducible services for Nix development environme
 
 ![Demo](./doc/demo.gif)
 
+## Motivation
+
+With `services-flake`, we aim to solve the following problems:
+
+- Run external services like databases, Redis, etc. natively across platforms.
+- Enable users to configure multiple instances of these external services.
+- Provide project-specific service configuration and data isolation.
+
 ## Getting Started
 
 See <https://community.flake.parts/services-flake/start>
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ With `services-flake`, we aim to solve the following problems:
 - Enable users to configure multiple instances of these external services.
 - Provide project-specific service configuration and data isolation.
 
+Consider a scenario where we are juggling two projects, located at `~/code/foo` and `~/code/bar`. The `foo` project integrates postgres and nginx, while `bar` encompasses postgres, pgAdmin, and kafka. It's crucial that the postgres data remains segregated across these projects. Additionally, the `bar` project is designed to facilitate multiple instances of postgres. Both projects are equipped with a flake app, streamlining the launch of their respective service stacks. Consequently, anyone using a Linux or macOS system can effortlessly clone either project and execute `nix run .#services` to activate the full suite of services without the need for manual configuration.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,6 +6,14 @@ Declarative, composable, and reproducible services for Nix development environme
...
 ![Demo](./doc/demo.gif)
 
+## Motivation
+
+With `services-flake`, we aim to solve the following problems:
+
+- Run external services like databases, Redis, etc. natively across platforms.
+- Enable users to configure multiple instances of these external services.
+- Provide project-specific service configuration and data isolation.
+
```
srid, line 16: I was thinking more like a narrative-telling. Here's one attempt ⤵️ 

```suggestion

Consider a scenario where we are juggling two projects, located at `~/code/foo` and `~/code/bar`. The `foo` project integrates postgres and nginx, while `bar` encompasses postgres, pgAdmin, and kafka. It's crucial that the postgres data remains segregated across these projects. Additionally, the `bar` project is designed to facilitate multiple instances of postgres. Both projects are equipped with a flake app, streamlining the launch of their respective service stacks. Consequently, anyone using a Linux or macOS system can effortlessly clone either project and execute `nix run .#services` to activate the full suite of services without the need for manual configuration.
```

doc/start.md
```diff
@@ -58,3 +58,7 @@ nix run
 [^how-default]: `default` is the name of the process group that is derived from `process-compose.<name>` in `perSystem.process-compose`.
 
 [^sample-config]: See the example configuration from the [getting started](https://f1bonacc1.github.io/process-compose/intro/) section of the process-compose docs.
+
+## See also
+
+If you are looking to replace docker and use `services-flake` in your dev environment, read about how we accomplished it at [Nammayatri](https://github.com/nammayatri/nammayatri): <https://nixos.asia/en/blog/replacing-docker-compose>
```
srid, line 64:
```suggestion
## Examples

- In [Nammayatri](https://github.com/nammayatri/nammayatri), services-flakes is used to run the local services stack (which used to be run with docker-compose). Read about it [in this blog post](https://nixos.asia/en/blog/replacing-docker-compose).
```

srid: Are there other public examples to add?

README.md
angerman: I actually think there is value in both. The bullet points already help quite well establish why I would want this, @srid explainarion adds a more concrete usecase ontop.

srid: Yea. We probably need a "Features" heading. 

Also, maybe we should create a comparison table as well? Comparing it to devenv and the like. @NorfairKing is good at this stuff. See examples:

- https://github.com/NorfairKing/sydtest?tab=readme-ov-file#features--comparison-to-similar-projects
- https://github.com/NorfairKing/feedback?tab=readme-ov-file#comparison-with-other-tools
- https://github.com/NorfairKing/really-safe-money?tab=readme-ov-file#comparison

PR #133, opened 2024-03-07
Improve speed of tests

srid:
Running `nixci` on my mac is extremely slow mainly due to grafana build. The slowness also translates to all of CI platforms.

Tests should run swiftly as possible. Our project is not complicated. If something's fucked in nixpkgs, we shouldn't have to get hit by slow build times.


shivaraj-bh:
We can improve the build time by disabling check for grafana instead.

srid:
Also less code is generally better. We don't want to maintain our own $package derivations. That's nixpkgs' job.

test/nix/pkgs.nix
```diff
@@ -0,0 +1,29 @@
...
+
+      # Required for elastic search
+      config.allowUnfree = true;
+
+      overlays = [
+        (self: super: lib.optionalAttrs super.stdenv.isDarwin {
+
+          # grafana is broken on aarch64-darwin, but works on older nixpkgs:
+          # https://github.com/NixOS/nixpkgs/issues/273998
+          grafana = (builtins.getFlake "github:nixos/nixpkgs/b604023e0a5549b65da3040a07d2beb29ac9fc63").legacyPackages.${system}.grafana;
```
srid, line 16: Just one line: we don't even have to add an extra flake input.

PR #135, opened 2024-03-07
Compare to similar tools

srid:
[**Rendered**](https://github.com/juspay/services-flake/blob/compare/README.md#comparison-with-other-tools)

- What other tools should we compare with?
- What other aspects should we mention?

This table is just a start, and not complete.


srid:
cc @roberth for ideas, especially in regards to the aspects. Currently, we have 'pure flakes', 'sharing' and 'flake apps' (devenv doesn't produce runnable packages, as far as I can tell).

image

srid:
@shivaraj-bh To substantiate our "share services across flakes", we probably need to document it first.

README.md
```diff
@@ -23,6 +23,20 @@ See <https://community.flake.parts/services-flake/start>
 
 List of supported services is available at https://community.flake.parts/services-flake/services
 
+## Comparison with similar approaches
+
+| | services-flake | [devenv](https://devenv.sh/) |
+| --- | --- | --- |
+| macOS support | ✔️  | ✔️  |
+| Pure Flakes | ✔️  | ❌[^1]  |
+| Share services across flakes | ✔️[^2]  | ❌  |
```
srid, line 32: This maybe inaccurate.

cf. https://devenv.sh/guides/using-with-flake-parts/#import-a-devenv-module

But I haven't used devenv enough to say if its module system works without any idiosyncratic shortcomings.

README.md
```diff
@@ -23,6 +23,20 @@ See <https://community.flake.parts/services-flake/start>
...
+| Services as flake apps | ✔️  | ❌[^3]  |
```
srid, line 33: From https://devenv.sh/guides/using-with-flake-parts/

> If you're familiar with the Nix language and ecosystem, devenv **can be used without the devenv CLI** by integrating into [Nix Flakes](https://www.tweag.io/blog/2020-05-25-flakes/) using [flake-parts](https://flake.parts/).

But the page doesn't explain how exactly one starts devenv services without using its CLI.

README.md
shivaraj-bh: Based on what I see, if you are able to `import` a module, you should be able to share services even using `devenv`.

shivaraj-bh: And I tried it out:

Here’s the `perSystem` block from my flake:

Screenshot 2024-03-08 at 11 43 16 AM


And this is the `devenv-module.nix`:

Screenshot 2024-03-08 at 11 43 43 AM


I do use `nix develop —impure` followed by `devenv up` to run them but in terms of sharing services, it works, i.e if you are willing to stick to devenv cli.

README.md
shivaraj-bh: In the flake I created [here](https://github.com/juspay/services-flake/pull/135/files#r1517209132), I explored the flake in nix repl, and these are the packages it exports:

```sh
nix-repl> packages.x86_64-linux.
packages.x86_64-linux.container-processes  packages.x86_64-linux.default
packages.x86_64-linux.container-shell      packages.x86_64-linux.devenv-up
```

shivaraj-bh: But I can’t seem to be able to `nix run` it:

```sh
❯ nix run .#devenv-up --impure
warning: Git tree '/home/shivaraj/oss/devenc' is dirty
trace: warning: getExe: Package "cowsay-3.7.0" does not have the meta.mainProgram attribute. We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold. If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away. Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar".
error: unable to execute '/nix/store/67lf41ig3sf8z4i6lnigfn4c0icyb830-devenv-up/bin/devenv-up': Not a directory
```

Probably it doesn’t have a bin output

shivaraj-bh: This works fine: `❯ nix build .#devenv-up —impure`

And produces a result whose contents are:

```sh
❯ cat /nix/store/67lf41ig3sf8z4i6lnigfn4c0icyb830-devenv-up
#!/nix/store/087167dfxal194pm54cmcbbxsfy3cjgn-bash-5.2p26/bin/bash


/nix/store/lb2b16q8kc3xynj27d25sahhq95ph43h-process-compose-0.88.0/bin/process-compose --config /nix/store/vhxxz7j5bjlyi0ggfx4dlcf9an61p09q-process-compose.yaml \
  --port ${PC_HTTP_PORT:-9999} \
  --tui=${PC_TUI_ENABLED:-1} \
  up "$@" &


if [[ ! -d "$DEVENV_STATE" ]]; then
  mkdir -p "$DEVENV_STATE"
fi

stop_up() {
  echo "Stopping processes..."
  kill -TERM $(cat "$DEVENV_STATE/devenv.pid")
  rm "$DEVENV_STATE/devenv.pid"
  wait

  echo "Processes stopped."
}

trap stop_up SIGINT SIGTERM

echo $! > "$DEVENV_STATE/devenv.pid"

wait


```

shivaraj-bh: But when I try to run it, `postgres` service never starts, it just times out and fails with:

```sh
❯ ./result
mkdir: cannot create directory ‘’: No such file or directory
./result: line 25: /devenv.pid: Permission denied
```

shivaraj-bh: But the same configuration works (including `postgres` starting successfully) when I do `nix develop —impure` and then `devenv up`

shivaraj-bh: If I run `./result` after removing `process.implementation = “process-compose”`, it uses [honcho by default ](https://devenv.sh/reference/options/#processimplementation), which seems to work fine.

I was able to execute queries using `psql`:

```sh
shivaraj=# create table id(id int);
CREATE TABLE
shivaraj=# \dt
        List of relations
 Schema | Name | Type  |  Owner
--------+------+-------+----------
 public | id   | table | shivaraj
(1 row)
```

shivaraj-bh: > But when I try to run it, `postgres` service never starts, it just times out and fails with:
> 
> ```shell
> ❯ ./result
> mkdir: cannot create directory ‘’: No such file or directory
> ./result: line 25: /devenv.pid: Permission denied
> ```

This is because of not exporting `DEVENV_STATE`, because I am no longer in devShell, I will have to do it manually:

```sh
 DEVENV_STATE=$(pwd) ./result
```

But even after that I couldn’t get it running using the `process-compose` implementation without being in the devShell

srid: In conclusion, we can confidently say that devenv doesn't support running services (nevermind individual process groups) as a flake app.

srid:
This is what it looks likes right now?

image

Anything that needs changing?

README.md
shivaraj-bh: >  we can't do the following, right?

Yes, hope I haven’t missed something. Let me give it one more shot to be certain, don’t want to make false claims because of missing some details.

README.md
shivaraj-bh: Looks like it, yes

shivaraj-bh:
IMO adding an example each for “Share services” and “Reusable services” would be useful.

shivaraj-bh:
Right, you had mentioned this earlier: https://github.com/juspay/services-flake/pull/135#issuecomment-1984404317

That’s something I can work on adding.

shivaraj-bh:
> @shivaraj-bh To substantiate our "share services across flakes", we probably need to document it first.

Here’s an example: https://github.com/juspay/services-flake/pull/137

Once we confirm on the example, I will also flesh out the a doc for this.

README.md
srid: @shivaraj-bh https://discord.com/channels/1036369714731036712/1036369758024650792/1215626814710620190

srid: I haven't gotten any clear and unambiguous from the devenv community, so let's just merge this. If they correct us, we can change the README obviously.

srid:
Merging. Let's make follow-up edits in separate PRs.

PR #137, opened 2024-03-09
docs(example): add share-services

shivaraj-bh:
Run services:
```sh
nix run "github:juspay/services-flake/share-services?dir=example/share-services"
```

Run integration-test:
```sh
nix run "github:juspay/services-flake/share-services?dir=example/share-services"#integration-test
```


srid (commented):
`feat(example)` or `docs(example`)? Because this PR doesn't add new features to services-flake.

So this share-services example is more of demonstration of how the process-compose-flake module system can be used to share services, right? I wonder if we can improve it in a couple of ways:

1. Have two separate flakes -- `share-services/common`, `share-services/foo` -- where foo imports services from common. The integration test can just remain in foo. I think this is more clear (re: structure) to the user. In fact, we might as well add a bar flake that imports common again.
2. Instead of running empty pg1 and r1, what if had something demo-worthy as well? Like example/simple that shows pgweb?

srid:
> 2\. Instead of running empty pg1 and r1, what if we had something equally demo-worthy as example/simple (which shows pgweb)?

Perhaps we can just clone `example/simple` and refactor it. Viz., put the northwind db service in `example/share-services/database/flake.nix`:

https://github.com/juspay/services-flake/blob/961bfd3ff8177fd0a0e625d7c6e37b9a11259fb0/example/simple/flake.nix#L30-L39

And then use it, for running pgweb, from `example/share-services/frontend/flake.nix`:

https://github.com/juspay/services-flake/blob/218fa6cc9a875c20def622e974fb4e0f391b3b6d/example/flake.nix#L49-L57

? That's one idea.

Interestingly, in future we can create another example to demonstrate creation of new services, by refactoring this pgweb into its own service (`services.pgweb.pgservice = "pg1"`).

shivaraj-bh:
@srid I have so far implemented [this](https://github.com/juspay/services-flake/pull/137/commits/a5bc86123a3b0f92f115f12962f8b440046f2f69). Although, I am not super pleased with the way [services.nix](https://github.com/juspay/services-flake/blob/a5bc86123a3b0f92f115f12962f8b440046f2f69/example/share-services/databases/services.nix) is being used between `databases` and `frontend`, any suggestions?

shivaraj-bh:
Its not a bug, data will still remain in pwd, only the socket will be created elsewhere. Hence, the options, `dataDir` and `socketDir` are different.

example/share-services/databases/services.nix
```diff
@@ -0,0 +1,27 @@
...
+    enable = true;
+    listen_addresses = "127.0.0.1";
+    initialDatabases = [
+      {
+        name = "sample";
+        schemas = [ "${inputs.northwind}/northwind.sql" ];
+      }
+    ];
+    # Required due to socket length being limited to 100 chars, see: https://github.com/juspay/services-flake/pull/77
+    socketDir = "$HOME/.services/postgres/pg1";
```
srid, line 13: Wouldn't this break if running two test builds in parallel?

Also: https://github.com/juspay/services-flake/issues/139#issuecomment-1988436374

shivaraj-bh:
Also, if we move the `northwind` DB to `share-services` example, would it make sense to have a much simpler example in `simple`?

example/share-services/databases/services.nix
shivaraj-bh: It will, yes! It should probably depend on a tmp path, rather than a pre-defined one. Gotta think about this one.

srid:
> Also, if we move the `northwind` DB to `share-services` example, would it make sense to have a much simpler example in `simple`?

Simple, as in ?

srid:
@shivaraj-bh These files seem out of place; were they added by accident?

image

srid:
n/m - that's just old code in the PR.

shivaraj-bh:
> > Also, if we move the `northwind` DB to `share-services` example, would it make sense to have a much simpler example in `simple`?
> 
> 
> 
> Simple, as in ?

n/m, I was thinking of just having ```
services.postgres.pg.enable = true;
services.redis.r1.enable = true;
```

But such simple cases are anyways covered in the "Getting started" section of the service documentation.

srid:
Good enough, we can always improve our examples later, and add more of them. I'll merge.

PR #143, opened 2024-03-12
feat(postgres): add connectionURI option

### LINKED ISSUE #141: postgres: add an option exposing connection URI
Instead of:

https://github.com/juspay/services-flake/blob/218fa6cc9a875c20def622e974fb4e0f391b3b6d/example/flake.nix#L54

we want to be able to:

```nix
{
  environment.PGWEB_DATABASE_URL = pgcfg.connectionURI { inherit dbName; };
  ...
}
```

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS

shivaraj-bh:
resolves #141

### DIFF
--- a/nix/postgres/default.nix
+++ b/nix/postgres/default.nix
@@ -56,6 +56,14 @@ in
       description = "The DB socket directory";
     };
 
+    # Based on: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS
+    connectionURI = lib.mkOption {
+      type = lib.types.functionTo lib.types.str;
+      readOnly = true;
+      default = dbName: "postgres://${config.listen_addresses}:${builtins.toString config.port}/${dbName}";
+      description = "A function that accepts database name and returns the [postgres connection URI](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS)";
+    };
+
     hbaConf =
       let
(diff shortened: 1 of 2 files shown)


nix/postgres/default.nix
```diff
@@ -56,6 +56,14 @@ in
       description = "The DB socket directory";
     };
 
+    # Based on: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS
+    connectionURI = lib.mkOption {
+      type = lib.types.functionTo lib.types.str;
+      readOnly = true;
+      default = dbName: "postgres://${config.listen_addresses}:${builtins.toString config.port}/${dbName}";
```
srid, line 63: I can imagine being able to pass additional arguments in the future. So the arguments should probably be an attrset.

See https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS

srid (commented):
This new option can now be used here:

https://github.com/juspay/services-flake/blob/8c25972cfb166f8c0718ea8e44859c8236a2fb89/example/share-services/pgweb/flake.nix#L37

nix/postgres/default.nix
shivaraj-bh: Here you go: https://github.com/juspay/services-flake/pull/146

PR #147, opened 2024-03-12
docs(postgres): init; specify unix-domain socket length limitation

shivaraj-bh:


doc/postgresql.md
```diff
@@ -0,0 +1,34 @@
...
+}
+```
+
+## Examples
+
+- Run postgres server initialised with a sample database and graphically interact with it using [pgweb](https://github.com/sosedoff/pgweb): <https://github.com/juspay/services-flake/tree/main/example/simple>
+
+## Gotchas
+
+### Unix-domain socket path is too long
```
srid, line 20:
```suggestion
{#socket-path}
### Unix-domain socket path is too long
```

doc/postgresql.md
```diff
@@ -0,0 +1,34 @@
...
+
+We already talk about this in the [data directory guide](https://community.flake.parts/services-flake/datadir#socket-path). In case of postgres, you can set `socketDir` while keeping the `dataDir` unchanged.
```
srid, line 22: Relative should work I believe?

```suggestion
We already talk about this in the [data directory guide](datadir#socket-path). In case of postgres, you can set `socketDir` while keeping the `dataDir` unchanged.
```

srid: And it should be relative path to the `.md` file (not the slug URL).

srid (approved):
LGTM, otherwise.

PR #150, opened 2024-03-13
docs: (readme) link to `datadir` doc from `cfp`; (index) add guide

shivaraj-bh:
- `Guide` page is not a child of `services-flake` in <https://community.flake.parts> page:

Screenshot 2024-03-13 at 3 54 20 PM

- Link to a more detailed datadir documentation in README


shivaraj-bh:
The misalignment of page should not happen in the future owing to: https://github.com/juspay/services-flake/pull/149

PR #154, opened 2024-03-18
fix(mysql)!: look for `*.sql` files in the top-level schema directory

shivaraj-bh:
* mysql service would assume that all the `*.sql` files in the directory, provided by `initialDatabases [{ schema = <directory>; ... }]`, exists in a folder named [mysql-databases](https://github.com/juspay/services-flake/blob/291e7be83aa0f24b8e40fc61df4a895827788707/nix/mysql.nix#L239-L241), after this change it would just look for all the `*.sql` files in the top-level schema directory.
* add tests

### DIFF
--- a/nix/mysql/default.nix
+++ b/nix/mysql/default.nix
@@ -216,7 +216,7 @@ in
 '';
 
             configureScript = pkgs.writeShellScriptBin "configure-mysql" ''
-              PATH="${lib.makeBinPath [config.package pkgs.coreutils]}:$PATH"
+              PATH="${lib.makeBinPath [config.package pkgs.coreutils pkgs.findutils]}:$PATH"
               set -euo pipefail
               ${envs} 
               ${lib.concatMapStrings (database: ''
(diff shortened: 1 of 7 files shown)


nix/mysql/default.nix
```diff
@@ -216,7 +216,7 @@ in
 '';
 
             configureScript = pkgs.writeShellScriptBin "configure-mysql" ''
-              PATH="${lib.makeBinPath [config.package pkgs.coreutils]}:$PATH"
+              PATH="${lib.makeBinPath [config.package pkgs.coreutils pkgs.findutils]}:$PATH"
```
srid, line 219: Any reason not to just use `writeShellApplication` and put these in `runtimeInputs`?

srid: We probably wanna use it anyway for shellcheck safety.

shivaraj-bh: > Any reason not to just use writeShellApplication and put these in runtimeInputs?

Nothing specific, I just kept the changes specific to the PR description.

shivaraj-bh: I can change it to `writeShellApplication` it shouldn’t be a problem. 

Or maybe, I can merge this one and create one PR that will change it in all the places.

srid: > Or maybe, I can merge this one and create one PR that will change it in all the places.

👍

shivaraj-bh: https://github.com/juspay/services-flake/pull/155

PR #156, opened 2024-03-20
docs: create placeholders

shivaraj-bh:


doc/examples.md
```diff
@@ -0,0 +1,4 @@
+# Examples
+
+- [ ] [[simple]]
+- [ ] [[share-services]]
```
srid, line 4:
```suggestion
- [ ] [[simple]]#
- [ ] [[share-services]]#
```

You also want to link `[[examples]]#` from somewhere. 

Make sure no dangling pages are shown.

srid: I don't think we need dedicates pages for each example. But `share-services.md` seems like a right candidate for actually documenting how to share them.

PR #157, opened 2024-03-21
docs(mysql): init

shivaraj-bh:


doc/mysql.md
```diff
@@ -1,6 +1,46 @@
...
+}
+```
+
+{#tips}
+
+## Tips & Tricks
+
+{#port}
+
+### Use a different port
```
srid, line 22: Just for posterity: A newline is not required between ID and heading.

PR #160, opened 2024-03-26
fix(postgres): empty `socketDir` by default

### LINKED ISSUE #139: postgres: Document socket length limitation and workaround
image

https://github.com/juspay/services-flake/pull/77

shivaraj-bh:
resolves #139 

the solution is implemented based on https://github.com/juspay/services-flake/issues/139#issuecomment-1998114104

An empty `socketDir` would mean that by default the `postgres` service will use TCP/IP instead of Unix-domain sockets.

### DIFF
--- a/doc/postgresql.md
+++ b/doc/postgresql.md
@@ -20,6 +20,9 @@
 {#socket-path}
 ### Unix-domain socket path is too long
 
+> [!warning]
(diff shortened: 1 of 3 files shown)


srid (approved):
Do the docs need a mention here?

https://community.flake.parts/services-flake/postgresql#socket-path

shivaraj-bh:
mentioned it as a `warning`

PR #163, opened 2024-04-10
Fix pgadmin scripts on MacOS

Javyre:
Hey, just a little fix for mac:
`readlink -m` is invalid on mac. Added coreutils as a runtime input for darwin.

Other scripts in the repo have a similar issue (using the -m flag for e.g.) but I am not using them atm so cannot QA very well.

### DIFF
--- a/nix/pgadmin.nix
+++ b/nix/pgadmin.nix
@@ -103,7 +103,11 @@ in
               let
                 setupScript = pkgs.writeShellApplication {
                   name = "setup-pgadmin";
-                  runtimeInputs = [ config.package ];
+                  runtimeInputs =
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
Thanks for the contribution! One small addition, could you reword the commit to something like: `fix(pgadmin): <commit-description>`. This has to be documented, but we are planning to auto-generate changelog from commits and this will be helpful.

shivaraj-bh:
I have fixed it one other service: https://github.com/juspay/services-flake/pull/164

We use `readlink -m` even in `cassandra` but the `startScript` there already has `coreutils` as a runtimeInput.

PR #164, opened 2024-04-10
fix(grafana): add `coreutils` as runtimeInput for the startScript

shivaraj-bh:
`readlink` shipped by default in macOS doesn't support `-m` flag

### DIFF
--- a/nix/grafana.nix
+++ b/nix/grafana.nix
@@ -65,7 +65,11 @@ in
             grafanaConfigIni = iniFormat.generate "defaults.ini" grafanaConfig;
             startScript = pkgs.writeShellApplication {
               name = "start-grafana";
-              runtimeInputs = [ config.package ];
+              runtimeInputs =
+                [ config.package ] ++
+                (lib.lists.optionals pkgs.stdenv.isDarwin [
+                  pkgs.coreutils
+                ]);
               text = ''
                 grafana server --config ${grafanaConfigIni} \
                                --homepath ${config.package}/share/grafana \


srid:
We have macOS CI; shouldn't `grafana_test.nix` have caught this?

shivaraj-bh:
Yes, I tried to debug this.

When the `nix flake check` happens on macOS, grafana doesn’t have any problem finding the data dir with `readlink -m` (hinting that it is somehow using the readlink from `coreutils`):

```sh
grafana-test> [gf1      ] logger=sqlstore t=2024-04-14T02:55:20.597021Z level=info msg="Creating SQLite database file" path=/private/tmp/nix-build-grafana-test.drv-0/data/gf1/grafana.db
```

But when I run the process-compose app with `nix run`, it chooses to use the readlink from macOS’ `/usr/bin`:

```sh
[gf1    ] readlink: illegal option -- m
[gf1    ] usage: readlink [-fn] [file ...]
```

As a result, it tries to default to the `/nix/store` path as its data directory:

```sh
[gf1    ] logger=sqlstore t=2024-04-14T08:36:59.551036+05:30 level=info msg="Creating SQLite database file" path=/nix/store/yjfc9zvy1i1dsm9jr4ixdh5fdf062agy-grafana-10.4.1/share/grafana/grafana.db
```

shivaraj-bh:
Looks like it is to do with `runCommand`:

```sh
nix-repl> :b legacyPackages.aarch64-darwin.runCommand "readlink-test" { } "${legacyPackages.aarch64-darwin.which}/bin/which readlink > $out”

This derivation produced the following outputs:
  out -> /nix/store/p50kx540nd6d73wqd8jik6n064ww34lx-readlink-test

❯ cat /nix/store/p50kx540nd6d73wqd8jik6n064ww34lx-readlink-test
/nix/store/vinxz6lkrilb72dkzr3ny02nnvql6z50-coreutils-9.4/bin/readlink
```

shivaraj-bh:
`runCommand`’s drv has `stdenv-darwin` as an input, which is where `coreutils` is coming from: 

```sh
"inputDrvs": {
      "/nix/store/32l6ffg3wc6vxnd6fdrjar7ys20k6faw-stdenv-darwin.drv": {
        "dynamicOutputs": {},
        "outputs": [
          "out"
        ]
      },
    }
```

PR #165, opened 2024-04-12
fix(postgres): fix pg_isready issue with empty listen_addresses

johnhampton:
This PR resolves the issue where `pg_isready` would fail when `socketDir` is set and `listen_addresses` is empty. `pg_isready` is now modified to use `socketDir` when it is available.

### DIFF
--- a/nix/postgres/default.nix
+++ b/nix/postgres/default.nix
@@ -323,7 +323,7 @@ in
                   '';
                 };
                 pg_isreadyArgs = [
-                  "-h ${config.listen_addresses}"
+                  (if config.socketDir != "" then "-h $(readlink -f \"${config.socketDir}\")" else "-h ${config.listen_addresses}")
                   "-p ${toString config.port}"
                   "-d template1"
                 ] ++ (lib.optional (config.superuser != null) "-U ${config.superuser}");
--- a/nix/postgres/postgres_test.nix
+++ b/nix/postgres/postgres_test.nix
@@ -16,8 +16,19 @@
       }
     ];
   };
+  services.postgres."pg3" = {
+    enable = true;
+    socketDir = "./test/new/socket/path2";
+    listen_addresses = "";
+    initialDatabases = [
+      {
+        name = "test-db";
+      }
+    ];
+  };
   # avoid both the processes trying to create `data` directory at the same time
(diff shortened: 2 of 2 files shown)


srid (commented):
Shoudn't this be done at a level close to the option def? Maybe:

https://github.com/juspay/services-flake/blob/e4d877059c7f8bc2a11324e52fa254e4b14070af/nix/postgres/default.nix#L182-L187

(I haven't looked the whole code, so I might be missing something)

shivaraj-bh:
> Shoudn't this be done at a level close to the option def?

I doubt that is possible here, because then we would expect `pg_isready` to automatically detect, which amongst `unix_socket_directories` or `listen_addresses` to choose from based on `postgresql.conf` but don’t think that is supported: https://www.postgresql.org/docs/current/app-pg-isready.html

shivaraj-bh:
@johnhampton could you also add a basic test `services.postgres.”pg3”`, that would test for empty `listen_addresses` and non-empty `socketDir`?

You can add the test here: <https://github.com/juspay/services-flake/blob/main/nix/postgres/postgres_test.nix>

johnhampton:
> @johnhampton could you also add a basic test `services.postgres.”pg3”`, that would test for empty `listen_addresses` and non-empty `socketDir`?
> 
> You can add the test here: https://github.com/juspay/services-flake/blob/main/nix/postgres/postgres_test.nix

@shivaraj-bh done.

johnhampton:
> Shoudn't this be done at a level close to the option def? Maybe:

@srid I'm not sure, I followed the code just above my change.

https://github.com/juspay/services-flake/blob/0922b13db8eff1ab8819e3202d2045a32ec719bf/nix/postgres/default.nix#L317-L322

shivaraj-bh:
> I doubt that is possible here, because then we would expect pg_isready to automatically detect, which amongst unix_socket_directories or listen_addresses to choose from based on postgresql.conf but don’t think that is supported: https://www.postgresql.org/docs/current/app-pg-isready.html

And primarily that is because the settings above (`postgresql.conf`) only applies to the server and not the client CLI utils.

I couldn’t find any documentation on the clients picking up configs from `postgresql.conf`, but did for the server: https://www.postgresql.org/docs/current/config-setting.html

shivaraj-bh (approved):
LGTM! Thanks

PR #170, opened 2024-04-23
fix: use absolute socket path in configureTimezones. this fixes #169

### LINKED ISSUE #169: Could not create unix socket lock file error in mysql service when timezones are imported
If `importTimeZones` is true or `default-time-zone` then the `configureTimezones` script will be executed during `startScript` initialization, if and only the dataDir does not exists.

The initialization fails with similar error:
```text
[mysql	] 2024-04-22T15:42:53.257499Z 0 [ERROR] Could not create unix socket lock file ./data/mysql/config.sock.lock.
```

The socket path used in `configureTimezones` script is relative and mysql fails to create it.

attilaersek:
The change fixes #169. Tested locally. To reproduce the original issue just try to spin up a local mysql service using `nixpkgs-old.legacyPackages.mysql57` package while `mysqld.default-time-zone` or `importTimeZones` is set.

### DIFF
--- a/nix/mysql/mysql_test.nix
+++ b/nix/mysql/mysql_test.nix
@@ -7,6 +7,12 @@
       CREATE USER bar;
     '';
   };
+  services.mysql.m3 = {
+    enable = true;
+    importTimeZones = true;
+    package = pkgs.mysql80;
+    settings.mysqld.port = 3309;
+  };
   services.mysql.m1.ensureUsers = [
     {
       name = "test_database";
@@ -14,8 +20,11 @@
       ensurePermissions = { "test_database.*" = "ALL PRIVILEGES"; };
     }
   ];
-  services.mysql.m2.enable = true;
-  services.mysql.m2.settings.mysqld.port = 3308;
+  services.mysql.m2 = {
+    enable = true;
+    importTimeZones = true;
+    settings.mysqld.port = 3308;
+  };
   settings.processes.test =
     {
       command = pkgs.writeShellApplication {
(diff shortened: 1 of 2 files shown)


shivaraj-bh (approved):
Looks good.

Could you also include a test for this by enabling `importTimeZones` in either `mysql.m1` or `mysql.m2` here: https://github.com/juspay/services-flake/blob/main/nix/mysql/mysql_test.nix?

nix/mysql/mysql_test.nix
```diff
@@ -7,6 +7,19 @@
       CREATE USER bar;
     '';
   };
+  services.mysql.m3 = {
+    enable = true;
+    importTimeZones = true;
+    package = pkgs.mysql80;
+    initialDatabases = [{ name = "test_dabase_with_tzdata"; }];
```
shivaraj-bh, line 14: I don’t see this being useful for the test below, is it needed?

attilaersek: no, actually not needed. removing it.

shivaraj-bh:
> If you prefer merging the test cases into an existing one, just let me know.

I don’t mind adding a newer instance of mysql, if that’s required. I think for testing `mysql80` it makes sense to add a newer mysql `m3`, but we could merge the `m4` into `m2` since `m2` is doesn’t test anything apart from starting mariadb on a different port.

attilaersek:
I've improved the tzdata tests cases to actually check whether the  `time_zone_name` table is correctly populated or not. By default this table does not contain any row.

attilaersek:
> > If you prefer merging the test cases into an existing one, just let me know.
> 
> I don’t mind adding a newer instance of mysql, if that’s required. I think for testing `mysql80` it makes sense to add a newer mysql `m3`, but we could merge the `m4` into `m2` since `m2` is doesn’t test anything apart from starting mariadb on a different port.

Sure! I'll merge m4 and m2 into one to simplify the code a bit.

shivaraj-bh:
Thanks! Everything looks good at a glance. Let me do a quick test on my machine and we are good to merge. 

Meanwhile, could you do one last thing. Can you squash your commits into 2 commits? one fixing the importTimezones and another one for adding tests? We are using commit history to generate the changelogs in the future release, so this will be very helpful.

attilaersek:
> Thanks! Everything looks good at a glance. Let me do a quick test on my machine and then we are good to merge.
> 
> Meanwhile, could you do one last thing. Can you squash your commits into 2 commits? one fixing the importTimezones and another one for adding tests? We are using commit history to generate the changelogs in the future releases, so this will be very helpful.

I just wanted to ask! 👍

shivaraj-bh:
Awesome! Thanks for the contribution. Merging it

PR #172, opened 2024-04-25
feat(mysql): allow configuring socketDir on mysql

attilaersek:
Introduced a new attribute `socketDir` to control the UDS location. If null, defaults to `dataDir` to maintain backward compatibility. Covered with tests.

### DIFF
--- a/nix/mysql/default.nix
+++ b/nix/mysql/default.nix
@@ -21,6 +21,12 @@ in
       description = "The mysql data directory";
     };
 
+    socketDir = lib.mkOption {
+      type = types.nullOr types.str;
+      default = config.dataDir;
+      description = "The mysql socket directory. If null, defaults to dataDir.";
+    };
+
     settings = lib.mkOption {
       type = format.type;
       default = { };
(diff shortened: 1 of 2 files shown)


attilaersek:
ci failure seems to be an unrelated flakiness:
```
Error: est> [m3 ] 2024-04-25T11:32:19.304870Z 0 [ERROR] [MY-011292] [Server] Plugin mysqlx reported: 'Preparation of I/O interfaces failed, X Protocol won't be accessible'
Error: est> [m3 ] 2024-04-25T11:32:19.304937Z 0 [ERROR] [MY-011300] [Server] Plugin mysqlx reported: 'Setup of socket: '/run/mysqld/mysqlx.sock' failed, can't create lock file /run/mysqld/mysqlx.sock.lock'
```
am i right?
could you please rerun the failed jobs?

attilaersek:
I merged the new m4 into the existing m2 to reduce resource pressure and I've also found out that mariadb was sensitive to the existence of the socket directory so changed the directory creation order.

shivaraj-bh:
We had a similar ask in `postgres`, see: https://github.com/juspay/services-flake/pull/77. This was introduced because of the [unix-socket length limitation](https://linux.die.net/man/7/unix). Is this the same reason to have this option here?

What happens in mysql if the socket length exceeds 100 or so characters? Does it fallback to using TCP/IP or does it crash the server?

I am asking this because, due to this behaviour we had decided to make `socketDir` in `postgres` empty by default, so that it uses TCP/IP; [PR](https://github.com/juspay/services-flake/pull/160).

If that’s not a concern with MySQL, it should make sense to have it set to `dataDir` by default.

attilaersek:
I'm not a mysql expert, I'm not sure if UDS can be disabled at all, but it would require a major change in the current initialization scripts.

attilaersek:
Let me know if you need help in future improvements! Appreciate your quick responses!

shivaraj-bh:
> Let me know if you need help in future improvements! Appreciate your quick responses!

Thanks! I will just quickly test it on my machine and then merge

shivaraj-bh:
Works great! Thanks for the contribution

Issue #90, opened 2024-02-03
Add Weaviate service


sharananurag998:
For AI projects in Juspay, we are using a vector database called [Weaviate](https://github.com/weaviate/weaviate). Need to add support for this in services-flake. 

[Nix package](https://search.nixos.org/packages?channel=23.11&from=0&size=50&sort=relevance&type=packages&query=weaviate)

- [ ] Expose options to configure the Weaviate server 
- [ ] Enable startiing `weaviate-server` in a process-compose window upon `nix run`

PR #195, opened 2024-05-25
Add Weaviate service

jedimahdi:
#90

### DIFF
--- a/nix/weaviate.nix
+++ b/nix/weaviate.nix
@@ -74,19 +74,17 @@ in
                   exec weaviate --scheme http --host ${config.host} --port ${toString config.port}
                 '';
               };
-
-              readyScript = pkgs.writeText "ready.py" ''
-                import weaviate
-                client = weaviate.connect_to_local(port=${toString config.port}, host="${config.host}")
-                client.close()
-              '';
             in
             {
               command = startScript;
 
               readiness_probe = {
-                exec.command = "${(pkgs.python3.withPackages (p: [ p.weaviate-client ]))}/bin/python ${readyScript}";
-                initial_delay_seconds = 2;
+                http_get = {
+                  host = config.host;
+                  port = config.port;
+                  path = "/v1/.well-known/ready";
+                };
+                initial_delay_seconds = 3;
                 period_seconds = 10;
                 timeout_seconds = 4;
                 success_threshold = 1;
--- a/nix/weaviate.nix
+++ b/nix/weaviate.nix
@@ -16,7 +16,7 @@ in
 
     host = lib.mkOption {
       type = types.nullOr types.str;
-      default = "0.0.0.0";
+      default = "127.0.0.1";
       description = ''
         The IP to listen on
       '';
--- a/nix/weaviate.nix
+++ b/nix/weaviate.nix
@@ -1,0 +1,101 @@
+{ pkgs, lib, name, config, ... }:
+let
+  inherit (lib) types;
+  format = pkgs.formats.json { };
+in
+{
+  options = {
+    enable = lib.mkEnableOption name;
+
+    package = lib.mkPackageOption pkgs "weaviate" { };
+
+    host = lib.mkOption {
+      type = types.nullOr types.str;
+      default = "0.0.0.0";
+      description = ''
+        The IP to listen on
+      '';
+      example = "0.0.0.0";
+    };
+
+    port = lib.mkOption {
+      type = types.port;
+      default = 8080;
+      description = ''
+        The port to listen on for connections
+      '';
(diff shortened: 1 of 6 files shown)


nix/weaviate_test.nix
```diff
@@ -0,0 +1,23 @@
+{ pkgs, config, ... }: {
+  services.weaviate."weaviate1" = {
+    enable = true;
+    envs = {
+      CLUSTER_ADVERTISE_ADDR = "127.0.0.1";
```
shivaraj-bh, line 5: Could you add a comment about why this env exists?

shivaraj-bh (approved):

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+in
+{
+  options = {
+    enable = lib.mkEnableOption name;
+
+    package = lib.mkPackageOption pkgs "weaviate" { };
+
+    dataDir = lib.mkOption {
+      type = types.str;
+      default = "./data";
```
srid, line 13: Shouldn't this be `./data/${name}`?

image

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+      description = "Path to the Weaviate data store";
+    };
+
+    host = lib.mkOption {
+      type = types.nullOr types.str;
+      default = "127.0.0.1";
+      description = ''
+        The IP to listen on
+      '';
```
srid, line 22: What happens if the user sets `host = null`? Is this behaviour documented in the option description?

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+
+    port = lib.mkOption {
+      type = types.port;
+      default = 8080;
+      description = ''
+        The port to listen on for connections
+      '';
+    };
+
+    envs = lib.mkOption {
```
srid, line 34: Why not call it `environment` to be in line with process-compose's option?

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+                else throw "Unrecognized type";
+
+              exports = (lib.mapAttrsToList (name: value: "export ${name}=${toStr value}") ({ "PERSISTENCE_DATA_PATH" = config.dataDir; }
+                // config.envs));
+
+              startScript = pkgs.writeShellApplication {
+                name = "start-weaviate";
+                runtimeInputs = [ config.package ];
+                text = ''
+                  ${lib.concatStringsSep "\n" exports}
```
srid, line 73: Instead of injecting environment variables in the shell script, why not set the process-compose `environment` option for the process? It is cleaner that way as well.

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+      description = ''
+        The port to listen on for connections
+      '';
+    };
+
+    envs = lib.mkOption {
+      type = types.attrsOf (types.oneOf [ types.str types.int types.bool (types.listOf types.str) ]);
+      default = { };
+      description = ''
+        Weaviate environment variables.
```
srid, line 38:
```suggestion
        Weaviate environment variables.
        
        See https://weaviate.io/developers/weaviate/config-refs/env-vars
```

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+      default = {
+        processes = {
+          "${name}" =
+            let
+              toStr = value:
+                if builtins.isString value then builtins.toJSON value
+                else if builtins.isBool value then (if value then "true" else "false")
+                else if builtins.isList value then builtins.toJSON (lib.concatStringsSep "," value)
+                else if builtins.isInt value then toString value
+                else throw "Unrecognized type";
```
srid, line 64: Why not:

```nix
let
  asStr = value:
    if builtins.isList value then lib.concatStringsSep "," value else value;
  toStr = value:
    builtins.toJSON (asStr value);
```

?

Then, instead of:

```nix
type = types.attrsOf (types.oneOf [ types.str types.int types.bool (types.listOf types.str) ]);
```

your type can be:

```nix
type = types.attrsOf (types.oneOf [ types.raw (types.listOf types.str) ]);
```

Also, is there a better way of defining this environment type?

nix/weaviate.nix
```diff
@@ -0,0 +1,102 @@
...
+                '';
+              };
+            in
+            {
+              command = startScript;
+
+              readiness_probe = {
+                http_get = {
+                  host = config.host;
+                  port = config.port;
```
srid, line 84:
```suggestion
                  inherit (config) host port;
```

srid (approved):
👍

Issue #107, opened 2024-02-21
Postgres: unify psql command execution in setup-script


shivaraj-bh:
@adrian-gierakowski I think it would be nicer if there is one function that executes psql commands so that we don't accidentally add a psql command in the future that doesn't use psqlUserArg. That can be done in a different PR, merging this.

_Originally posted by @shivaraj-bh in https://github.com/juspay/services-flake/pull/106#issuecomment-1957537267_

shivaraj-bh:
> or simply a bash snippet we can interpolate into the script

This sounds good!

PR #209, opened 2024-06-05
fix(postgres): stop init on error in `sql` scripts

### LINKED ISSUE #207: Errors in postgres `initialDatabases` schema are ignored
Using,

```
          services.postgres."pg" = {
            enable = true;
            initialDatabases = [
              {
                name = "ab_v2";
                schemas = [ ./scripts/db.sql ];
              }
            ];
          };
```

in an internal project (Infra team), gives:

image

Note the `ERROR` in the pg-init process. Yet it completed successfully, causing postgres to start. I'd expect the whole thing to fail.

When I scroll up, I can see a cascade of errors actually:

image

shivaraj-bh:
resolves #207 

From psql docs (https://www.postgresql.org/docs/current/app-psql.html):

> psql returns 0 to the shell if it finished normally, 1 if a fatal
error of its own occurs (e.g., out of memory, file not found), 2 if the connection to the server went bad and the session was not interactive, and 3 if an error occurred in a script and the variable ON_ERROR_STOP was set.

Also resolves https://github.com/juspay/services-flake/issues/107

### DIFF
--- a/nix/postgres/setup-script.nix
+++ b/nix/postgres/setup-script.nix
@@ -1,12 +1,11 @@
 { config, pkgs, lib }:
 let
-  psqlUserArg = lib.optionalString (config.superuser != null) "-U ${config.superuser}";
   setupInitialSchema = dbName: schema: ''
     echo "Applying database schema on ${dbName}"
     if [ -f "${schema}" ]
     then
       echo "Running file ${schema}"
-      awk 'NF' "${schema}" | psql ${psqlUserArg} -d ${dbName}
+      awk 'NF' "${schema}" | psql_with_args -d ${dbName}
     elif [ -d "${schema}" ]
     then
       # Read sql files in version order. Apply one file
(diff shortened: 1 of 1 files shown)


nix/postgres/setup-script.nix
```diff
@@ -79,7 +83,10 @@ in
   name = "setup-postgres";
   runtimeInputs = with pkgs; [ config.package coreutils gnugrep gawk ];
   text = ''
-    set -euo pipefail
+    set -x
+    function psql_with_args() {
+      psql ${psqlUserArg} -v "ON_ERROR_STOP=1" "$@"
+    }
```
srid, line 89: Add a comment for this function. 

Also, any reason not to inline (or move to a `let` inside of `text =`) `psqlUserArg` here? Furthermore, the `ON_ERROR_STOP` seems redundant here as `psqlUserArg` already has it above.

srid (approved):

Issue #210, opened 2024-06-06
Allow configurable dashboard providers in Grafana


alexpearce:
Similarly to datasources, it would be nice to be able to configure [dashboard providers](https://grafana.com/tutorials/provision-dashboards-and-data-sources/#provision-a-dashboard) in the Grafana service.

(PR incoming!)

PR #211, opened 2024-06-06
feat(grafana): add providers configuration

alexpearce:
Addresses juspay/services-flake#210.

I've tested this by hand with configuration:

```nix
services.grafana."grafana" = {
  enable = true;
  providers = [
    {
      name = "Databases";
      type = "file";
      options = {
        path = ./dashboards;
        foldersFromFilesStructure = true;
      };
    }
  ];
};
```

I haven't looked into it, but it might be possible to extend the existing Grafana test to check that a folder of dashboards is loaded.

### DIFF
--- a/nix/grafana.nix
+++ b/nix/grafana.nix
@@ -51,7 +51,11 @@ in
 
     datasources = lib.mkOption {
       type = types.listOf yamlFormat.type;
-      description = "List of data sources to configure.";
+      description = ''
+        List of data sources to configure.
+
+        See https://grafana.com/docs/grafana/latest/administration/provisioning/#data-sources for the schema.
+      '';
       default = [ ];
       example = ''
         [
@@ -71,6 +75,28 @@ in
       example = ''
         [
           { name = "Tempo"; }
+        ]
+      '';
+    };
+
+    providers = lib.mkOption {
+      type = types.listOf yamlFormat.type;
+      description = ''
+        List of dashboard providers to configure.
+
+        See https://grafana.com/docs/grafana/latest/administration/provisioning/#dashboards for the schema.
+      '';
+      default = [ ];
+      example = ''
+        [
+          {
+            name = "Databases";
+            type = "file";
+            options = {
+              path = ./dashboards;
+              foldersFromFilesStructure = true;
+            };
+          }
         ]
       '';
     };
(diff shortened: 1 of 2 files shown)


nix/grafana.nix
```diff
@@ -75,6 +75,24 @@ in
       '';
     };
 
+    providers = lib.mkOption {
+      type = types.listOf yamlFormat.type;
+      description = "List of dashboard providers to configure.";
```
shivaraj-bh, line 80: how about linking <https://grafana.com/docs/grafana/latest/administration/provisioning/#dashboards> in the description? The user will have idea of how the schema looks like

shivaraj-bh (commented):
It would be nice to also see a test for this option!

nix/grafana.nix
alexpearce: Good idea. I'll add a similar link for the datasources schema.

alexpearce:
I'm not sure how best to run the tests locally. `test/flake.nix` refers `services-flake` as `juspay/services-flake` so it won't pick up changes in my local repo.

shivaraj-bh:
> I'm not sure how best to run the tests locally. `test/flake.nix` refers `services-flake` as `juspay/services-flake` so it won't pick up changes in my local repo.

Once in the devShell you can run `just test grafana` to run the test locally

alexpearce:
Nice, thanks!

I had to modify the tests to use `nixos-24.05` to work around the Grafana-on-macOS issue, but with my latest commit the tests pass.

nix/grafana_test.nix
```diff
@@ -22,8 +76,12 @@
             ADMIN=${cfg.extraConf.security.admin_user}
             PASSWORD=${cfg.extraConf.security.admin_password}
             ROOT_URL="${cfg.protocol}://${cfg.domain}:${builtins.toString cfg.http_port}";
+            # The admin user can authenticate against the running service.
             curl -sSfN -u $ADMIN:$PASSWORD $ROOT_URL/api/org/users -i
             curl -sSfN -u $ADMIN:$PASSWORD $ROOT_URL/api/org/users | grep admin\@localhost
+            # The dashboard provisioner was used to create a dashboard.
+            curl -sSfN -u $ADMIN:$PASSWORD $ROOT_URL/api/dashboards/uid/${dashboardUid} -i
+            curl -sSfN -u $ADMIN:$PASSWORD $ROOT_URL/api/dashboards/uid/${dashboardUid} | grep '"title":${dashboardTitle}'
```
shivaraj-bh, line 84:
```suggestion
            curl -sSfN -u $ADMIN:$PASSWORD $ROOT_URL/api/dashboards/uid/${dashboardUid} | grep '"title”:"${dashboardTitle}"'
```

`grep` is failing in Linux because `dashboardTitle` is expected to be in `””`.

alexpearce: Absolutely!

shivaraj-bh:
Thanks @alexpearce

PR #218, opened 2024-06-11
feat(ollama): init

shivaraj-bh:
ported from https://github.com/shivaraj-bh/ollama-flake/blob/main/services/ollama.nix

TODO

- [x] Add documentation that includes examples for using accelerators like `cuda` and `rocm`

### DIFF
--- a/doc/ollama.md
+++ b/doc/ollama.md
@@ -1,0 +1,40 @@
+# Ollama
+
+[Ollama](https://github.com/ollama/ollama) enables you to get up and running with Llama 3, Mistral, Gemma, and other large language models.
+
+## Getting Started
+
+```nix
+# In `perSystem.process-compose.<name>`
+{
+  services.ollama."ollama1".enable = true;
+}
+```
+
+## Acceleration
+
+By default Ollama uses the CPU for inference. To enable GPU acceleration:
+
+### Cuda
+
+```nix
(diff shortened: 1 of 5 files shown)


srid:
@shivaraj-bh 

https://x.com/nixos_asia/status/1800855562072322052

srid:
@shivaraj-bh 

https://x.com/nixos_asia/status/1800855562072322052

Is there a `nix run` one-liner I can use to try out the demo in the tweet without having to hand-roll a flake? Perhaps an `./example/llm` is useful. I wanted to retweet that to my profile giving the `nix run ...` invocation for readers.

shivaraj-bh:
@srid here you go: https://github.com/juspay/services-flake/commit/82ff0e98761facd66de04672adb731e3f2d2d627

srid:
This is what I see when trying to run it:

```
❯ nix run "github:juspay/services-flake?dir=example/llm"
error:
       … while updating the lock file of flake 'github:juspay/services-flake/76f6081f51d8884b7d3196dbb45e8f8a11417a53?dir=example/llm'

       error: cannot write modified lock file of flake 'github:juspay/services-flake?dir=example/llm' (use '--no-write-lock-file' to ignore)
```

Also, we should build all example in CI. See #221

PR #219, opened 2024-06-11
feat(open-webui): init

shivaraj-bh:
ported from: https://github.com/shivaraj-bh/ollama-flake/blob/main/services/open-webui.nix

This was also recently upstreamed to [nixpkgs](https://github.com/NixOS/nixpkgs/tree/master): https://github.com/NixOS/nixpkgs/pull/316248

TODO:

- [x] Document how to use it with [ollama](https://github.com/juspay/services-flake/pull/218) service
- [x] Use `open-browser` process in example and documentation

### DIFF
--- a/doc/open-webui.md
+++ b/doc/open-webui.md
@@ -1,0 +1,68 @@
+# Open WebUI
+
+[Open WebUI](https://github.com/open-webui/open-webui) is a user-friendly WebUI for LLMs. It supports various LLM runners, including [[ollama]] and OpenAI-compatible APIs.
+
+{#start}
+## Getting Started
+
+```nix
+# In `perSystem.process-compose.<name>`
+{
+  services.open-webui."open-webui1".enable = true;
+}
+```
+
+## Examples
+
+{#ollama}
+### Open WebUI with ollama backend
+
+```nix
+{
+  services = {
+    # Backend service to perform inference on LLM models
+    ollama."ollama1" = {
+      enable = true;
+      # The models are usually huge, downloading them in every project directory can lead to a lot of duplication
+      dataDir = "$HOME/.services-flake/ollama1";
+      models = [ "llama2-uncensored" ];
+    };
+    # Get ChatGPT like UI, but open-source, with Open WebUI
+    open-webui."open-webui1" = {
+      enable = true;
+      environment =
+        let
+          inherit (pc.config.services.ollama.ollama1) host port;
+        in
+        {
+          OLLAMA_API_BASE_URL = "http://${host}:${toString port}";
+          WEBUI_AUTH = "False";
+        };
+    };
+  };
+  # Start the Open WebUI service after the Ollama service has finished initializing and loading the models
+  settings.processes.open-webui1.depends_on.ollama1-models.condition = "process_completed_successfully";
+}
+```
+
+See [[ollama]] for more customisation of the backend.
+
+{#browser}
+## Open browser on startup
+
+```nix
+{
+  services.open-webui."open-webui1".enable = true;
+  # Open the browser after the Open WebUI service has started
+  settings.processes.open-browser = {
+    command =
+      let
+        inherit (pc.config.services.open-webui.open-webui1) host port;
+        opener = if pkgs.stdenv.isDarwin then "open" else lib.getExe' pkgs.xdg-utils "xdg-open";
+        url = "http://${host}:${toString port}";
+      in
+      "${opener} ${url}";
+    depends_on.open-webui1.condition = "process_healthy";
+  };
+}
+```
--- a/example/llm/flake.nix
+++ b/example/llm/flake.nix
@@ -13,13 +13,45 @@
       inputs.process-compose-flake.flakeModule
     ];
     perSystem = { self', pkgs, lib, ... }: {
-      process-compose."default" = {
+      process-compose."default" = pc: {
         imports = [
           inputs.services-flake.processComposeModules.default
         ];
-        services.ollama."ollama1" = {
-          enable = true;
-          models = [ "llama2-uncensored" ];
+        services = {
+          # Backend service to perform inference on LLM models
+          ollama."ollama1" = {
+            enable = true;
+            # The models are usually huge, downloading them in every project directory can lead to a lot of duplication
+            dataDir = "$HOME/.services-flake/ollama1";
+            models = [ "llama2-uncensored" ];
+          };
+          # Get ChatGPT like UI, but open-source, with Open WebUI
+          open-webui."open-webui1" = {
+            enable = true;
+            environment =
+              let
+                inherit (pc.config.services.ollama.ollama1) host port;
+              in
+              {
+                OLLAMA_API_BASE_URL = "http://${host}:${toString port}";
+                WEBUI_AUTH = "False";
+              };
+          };
+        };
+
+        # Start the Open WebUI service after the Ollama service has finished initializing and loading the models
+        settings.processes.open-webui1.depends_on.ollama1-models.condition = "process_completed_successfully";
+
+        # Open the browser after the Open WebUI service has started
+        settings.processes.open-browser = {
+          command =
+            let
+              inherit (pc.config.services.open-webui.open-webui1) host port;
+              opener = if pkgs.stdenv.isDarwin then "open" else lib.getExe' pkgs.xdg-utils "xdg-open";
+              url = "http://${host}:${toString port}";
+            in
+            "${opener} ${url}";
+          depends_on.open-webui1.condition = "process_healthy";
         };
       };
     };
--- a/nix/open-webui.nix
+++ b/nix/open-webui.nix
@@ -1,0 +1,107 @@
+# Based on: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/open-webui.nix
+{ pkgs, lib, name, config, ... }:
+let
+  inherit (lib) types;
+in
+{
+  options = {
+    enable = lib.mkEnableOption "Open-WebUI server";
+    package = lib.mkPackageOption pkgs "open-webui" { };
+
+    dataDir = lib.mkOption {
+      type = types.str;
+      default = "./data/${name}";
+      description = "The Open-WebUI data directory";
+    };
+
+    host = lib.mkOption {
+      type = types.str;
+      default = "127.0.0.1";
+      example = "0.0.0.0";
+      description = ''
+        The host address which the Open-WebUI server HTTP interface listens to.
+      '';
+    };
+
+    port = lib.mkOption {
+      type = types.port;
+      default = 1111;
+      example = 11111;
+      description = ''
+        Which port the Open-WebUI server listens to.
+      '';
+    };
+
+    environment = lib.mkOption {
+      type = types.attrsOf types.str;
+      default = {
+        SCARF_NO_ANALYTICS = "True";
+        DO_NOT_TRACK = "True";
+        ANONYMIZED_TELEMETRY = "False";
+      };
+      example = ''
+        {
+          OLLAMA_API_BASE_URL = "http://127.0.0.1:11434";
+          # Disable authentication
+          WEBUI_AUTH = "False";
+        }
+      '';
+      description = "Extra environment variables for Open-WebUI";
+    };
+
+    outputs.settings = lib.mkOption {
+      type = types.deferredModule;
+      internal = true;
+      readOnly = true;
+      default = {
+        processes = {
+          "${name}" =
+            let
+              setupStateDirs = lib.concatMapStrings
+                (stateDir:
+                  ''
+                    if [ ! -d "''$${stateDir}" ]; then
+                      mkdir -p "''$${stateDir}"
+                    fi
+
+                    ${stateDir}=$(readlink -f "''$${stateDir}")
+
+                    export ${stateDir}
+                  '') [ "DATA_DIR" "STATIC_DIR" "HF_HOME" "SENTENCE_TRANSFORMERS_HOME" ];
+            in
+
+            {
+              environment = {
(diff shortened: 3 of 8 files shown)


nix/open-webui.nix
```diff
@@ -0,0 +1,95 @@
...
+              text = ''
+                set -x
+                if [ ! -d ${config.dataDir} ]; then
+                  mkdir -p ${config.dataDir} 
+                fi
+
+                DATA_DIR=$(readlink -f ${config.dataDir})
+                STATIC_DIR=$DATA_DIR
+                HF_HOME=$DATA_DIR
+                SENTENCE_TRANSFORMERS_HOME=$DATA_DIR
```
drupol, line 72: Is there some kind of attribute (a bit like with systemd) where we can define the environment variables?

drupol: Apparently it is possible, see: https://github.com/juspay/services-flake/blob/c9e553879a792b9a746f0b8395c9dbffd58d8342/example/share-services/pgweb/flake.nix#L32C1-L32C118

How about moving the environment variable out of the shell script and move them in there?

drupol: excellent now!

shivaraj-bh: Thanks @drupol! Did it over here: https://github.com/juspay/services-flake/pull/219/commits/72480e26c051930b449b6bb6c3b88e3b4161fa48

Now it is similar to what `systemd` allows.

nix/open-webui.nix
```diff
@@ -0,0 +1,106 @@
...
+                SENTENCE_TRANSFORMERS_HOME = config.dataDir;
+              } // config.environment;
+
+              command = pkgs.writeShellApplication {
+                name = "open-webui-wrapper";
+                runtimeInputs = [ pkgs.open-webui ];
+                text = ''
+                  ${setupStateDirs}
+
+                  open-webui serve --host ${config.host} --port ${builtins.toString config.port}
```
drupol, line 87:
```suggestion
                  ${lib.getExe config.package} serve --host ${config.host} --port ${builtins.toString config.port}
```

nix/open-webui.nix
```diff
@@ -0,0 +1,106 @@
...
+              environment = {
+                DATA_DIR = config.dataDir;
+                STATIC_DIR = config.dataDir;
+                HF_HOME = config.dataDir;
+                SENTENCE_TRANSFORMERS_HOME = config.dataDir;
+              } // config.environment;
+
+              command = pkgs.writeShellApplication {
+                name = "open-webui-wrapper";
+                runtimeInputs = [ pkgs.open-webui ];
```
drupol, line 83:
```suggestion
                runtimeInputs = [ config.package ];
```

nix/open-webui.nix
shivaraj-bh: In `writeShellApplication` we pass the runtime dependencies using `runtimeInputs` and that’s how we can expect the executable to be present in the environment. Do you see any problems in doing so?

drupol: If by any chance, there's another `open-webui` in the path, this might be a problem indeed. Usually it's safer to do as advised.

srid: > If by any chance, there's another `open-webui` in the path, this might be a problem indeed. 

Is it a problem in practice? Doesn't `writeShellApplication` prepend the runtimeInputs to `$PATH` anyway?

> This pattern is also used in `nixpkgs`.

Could you provide an example?

If we are using `lib.getExe`, we might as well remove `runtimeInputs` entirely.

drupol: I don't know by heart, ...but... just a simple search in `nixos/modules/services`:

image

Is it enough ? :)

srid: But none of them use `writeShellApplication`? 

I can see why in systemd's `ExecStart` we have to use `lib.getExe`, since there's no `runtimeInputs` that manipulates `$PATH` before running the script/command anyway.

drupol: That's true indeed, I overlooked that detail. In that case, I let you decide what to do here since my argument isn't totally true.

srid: I personally prefer `runtimeInputs` and not using `lib.getExe` in `writeShellApplication.

If we are however using `lib.getExe`, I wouldn't use `runtimeInputs`. 

The latter is probably safer, since it is possible to accidentally use the wrong package in `runtimeInputs` with `text` using an executable that is already in user's path.

drupol: As you prefer...

shivaraj-bh: Going with the safer option then.

doc/open-webui.md
```diff
@@ -0,0 +1,47 @@
...
+## Examples
+
+{#ollama}
+### Open WebUI with ollama backend
+
+```nix
+{
+  services.open-webui."ow1" = {
+    enable = true;
+    environment.OLLAMA_API_BASE_URL = "http://localhost:11434";
```
srid, line 24: We should try to avoid hardcoding wherever we can (it can also be self-documenting):

```suggestion
    environment.OLLAMA_API_BASE_URL = let ollama = config.services.ollama.ollama1; in "http://${cfg.output.settings.environment.OLLAMA_HOST}";
```

(Untested code ^)

https://github.com/juspay/services-flake/blob/76f6081f51d8884b7d3196dbb45e8f8a11417a53/nix/ollama.nix#L110

doc/open-webui.md
```diff
@@ -0,0 +1,47 @@
...
+
+```nix
+{
+  services.open-webui."ow1".enable = true;
+  settings.processes.open-browser = {
+    command = pkgs.writeShellApplication {
+      name = "open-browser";
+      runtimeInputs = lib.optionals pkgs.stdenv.isLinux [ pkgs.xdg-utils ] [ ];
+      text = ''
+        ${ if pkgs.stdenv.isLinux then "xdg-open http://127.0.0.1:1111" else "" }
```
srid, line 41: Same comment about hardcoding here ^

doc/open-webui.md
```diff
@@ -0,0 +1,47 @@
...
+# In `perSystem.process-compose.<name>`
+{
+  services.open-webui."ow1".enable = true;
+}
+```
+
+## Examples
+
+{#ollama}
+### Open WebUI with ollama backend
```
srid, line 18: This section should link to `ollama.md` in some fashion.

doc/open-webui.md
```diff
@@ -0,0 +1,47 @@
+# Open WebUI
+
+[Open WebUI](https://github.com/open-webui/open-webui) is a user-friendly WebUI for LLMs (Formerly Ollama WebUI).
```
srid, line 3:
```suggestion
[Open WebUI](https://github.com/open-webui/open-webui) is a user-friendly WebUI for LLMs. It supports various LLM runners, including [[ollama]] and OpenAI-compatible APIs.
```

example/llm/flake.nix
```diff
@@ -43,10 +47,14 @@
...
-            text = ''
-              ${ if pkgs.stdenv.isLinux then "xdg-open http://127.0.0.1:1111" else "" }
-              ${ if pkgs.stdenv.isDarwin then "open http://127.0.0.1:1111" else "" }
-            '';
+            text =
+              let
+                inherit (pc.config.services.open-webui.open-webui1) host port;
+              in
+              ''
+                ${ if pkgs.stdenv.isLinux then "xdg-open http://${host}:${toString port}" else "" }
```
drupol, line 55: You could use `lib.optionalString`, so you avoid writing the `else` clause.

shivaraj-bh:
Ready for preview:

```sh
nix run "github:juspay/services-flake/open-webui?dir=example/llm"
```

srid:
> Ready for preview:

I get the following on mac:

```
       error: The option `perSystem.aarch64-darwin.process-compose.default.services.open-webui' does not exist. Definition values:
       - In `/nix/store/7a9x52qf14dkvnhj4219jd0i5l9yhdj3-source/example/llm/flake.nix, via option perSystem':
           {
             open-webui1 = {
               enable = true;
               environment = {
                 OLLAMA_API_BASE_URL = "http://127.0.0.1:11434";
```

doc/open-webui.md
srid: @shivaraj-bh This conversation is marked as resolved but the diff still shows `"http://localhost:11434"` for me?

example/llm/flake.nix
```diff
@@ -13,13 +13,49 @@
...
+          command = pkgs.writeShellApplication {
+            name = "open-browser";
+            text =
+              let
+                inherit (pc.config.services.open-webui.open-webui1) host port;
+              in
+              ''
+                ${ lib.optionalString pkgs.stdenv.isLinux "${lib.getExe' pkgs.xdg-utils "xdg-open"} http://${host}:${toString port}"}
+                ${ lib.optionalString pkgs.stdenv.isDarwin "open http://${host}:${toString port}" }
+              '';
```
drupol, line 56:
```suggestion
            text =
              let
                inherit (pc.config.services.open-webui.open-webui1) host port;
              in
                lib.optionalString pkgs.stdenv.isLinux "${lib.getExe' pkgs.xdg-utils "xdg-open"} http://${host}:${toString port}"
                +
                lib.optionalString pkgs.stdenv.isDarwin "open http://${host}:${toString port}"
```

shivaraj-bh:
> > Ready for preview:
> 
> I get the following on mac:
> 
> ```
>        error: The option `perSystem.aarch64-darwin.process-compose.default.services.open-webui' does not exist. Definition values:
>        - In `/nix/store/7a9x52qf14dkvnhj4219jd0i5l9yhdj3-source/example/llm/flake.nix, via option perSystem':
>            {
>              open-webui1 = {
>                enable = true;
>                environment = {
>                  OLLAMA_API_BASE_URL = "http://127.0.0.1:11434";
> ```

Ah sorry (I was just checking with `just ex-llm` in devShell), this should work:

```sh
 nix run "github:juspay/services-flake/open-webui?dir=example/llm" --refresh --override-input services-flake github:juspay/services-flake/open-webui
```

example/llm/flake.nix
```diff
@@ -13,13 +13,49 @@
...
+          };
```
srid, line 57: Why not simply:

```nix
{
  command = 
    let 
      opener = if pkgs.stdenv.isDarwin then "open" else lib.getExe' pkgs.xdg-utils "xdg-open";
      url = "http://${host}:${toString port}";
    in "${opener} ${url}";
}
```

shivaraj-bh: Nice

example/llm/flake.nix
```diff
@@ -13,13 +13,45 @@
...
+        services = {
+          # Backend service to perform inference on LLM models
+          ollama."ollama1" = {
+            enable = true;
+            # The models are usually huge, downloading them in every project directory can lead to a lot of duplication
+            dataDir = "$HOME/.services-flake/ollama1";
+            models = [ "llama2-uncensored" ];
+          };
+          # Get ChatGPT like UI, but open-source, with Open WebUI
+          open-webui."open-webui1" = {
```
drupol, line 29: How about renaming `open-webui1` to `instance1` ? I found it confusing as it is now... don't you?

srid:
The processes are running in two different namespaces (`default` and `ollama1`) ... shouldn't they be under the same namespace?

image

example/llm/flake.nix
shivaraj-bh: Yes, I could go with that. Once I add the `namespace` setting to the `open-webui` service that should indicate that the current process will be `instance1` under `open-webui` namespace.

shivaraj-bh:
> The processes are running in two different namespaces (`default` and `ollama1`) ... shouldn't they be under the same namespace?
> 
> image

In our current design, every service belongs to its own namespace. We could manually modify this, if that’s what we desire here.

srid:
Fantastic, worked on mac!

image

image

srid:
> In our current design, every service belongs to its own namespace.

I don't remember what we do. But this doesn't seem to be the case for the llm example because only the `ollama1` service is running in its own namespace (`ollama`). Not the `open-webui1` service, as it is running in the `default` namespace.

shivaraj-bh:
> > In our current design, every service belongs to its own namespace.
> 
> I don't remember what we do. But this doesn't seem to be the case for the llm example because only the `ollama1` service is running in its own namespace (`ollama1`). Not the `open-webui1` service, as it is running in the `default` namespace.

Fixed it in the recent commit

srid:
> Fixed it in the recent commit

Re-ran `nix run` and everything still works (re-using the already downloaded models). This will be a cool video demo (especially if it finishes with writing prompts in the browser app).

example/llm/flake.nix
srid: I don't know, I quite like it the way it is.  "ollama1-models" is more clear to me than "instance1-models" for instance.

srid (commented):
LGTM, once the hardcoded strings in the docs are addressed.

example/llm/flake.nix
shivaraj-bh: I was trying it out, it does make sense to keep it as it is, for readability. The line adding `depends_on` relation between open-webui and ollama process would look like:
```nix
{
  # Start the Open WebUI service after the Ollama service has finished initializing and loading the models
  settings.processes.instance1.depends_on.ollama1-models.condition = "process_completed_successfully";
}
```

vs

(Current)
```nix
{
  # Start the Open WebUI service after the Ollama service has finished initializing and loading the models
  settings.processes.open-webui1.depends_on.ollama1-models.condition = "process_completed_successfully";
}
```

shivaraj-bh:
Thanks for all the feedback @srid and @drupol! ❤️

PR #241, opened 2024-06-19
searxng: init

drupol:
This PR:

- [x] Add the `searxng` service

Test with:

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
    flake-parts.url = "github:hercules-ci/flake-parts";
    systems.url = "github:nix-systems/default";
    process-compose-flake.url = "github:Platonic-Systems/process-compose-flake";
    services-flake.url = "github:juspay/services-flake";
  };
  outputs =
    inputs:
    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
      systems = import inputs.systems;
      imports = [ inputs.process-compose-flake.flakeModule ];
      perSystem =
        { pkgs, lib, ... }:
        {
          # `process-compose.foo` will add a flake package output called "foo".
          # Therefore, this will add a default package that you can build using
          # `nix build` and run using `nix run`.
          process-compose."default" =
            pc:
            let
              inherit (pc.config.services.searxng.searxng) host port;
            in
            {
              imports = [ inputs.services-flake.processComposeModules.default ];

              services.searxng."searxng" = {
                enable = true;
                settings = {
                  use_default_settings = true;
                  server.secret_key = "secret";
                  server.limiter = false;
                };
              };

              # Open the browser after the Open WebUI service has started
              settings.processes.open-browser = {
                command =
                  let
                    opener = if pkgs.stdenv.isDarwin then "open" else lib.getExe' pkgs.xdg-utils "xdg-open";
                    url = "http://${host}:${toString port}";
                  in
                  "${opener} ${url}";
                depends_on.searxng.condition = "process_healthy";
              };
            };
        };
    };
}
```

### DIFF
--- a/doc/searxng.md
+++ b/doc/searxng.md
@@ -1,0 +1,20 @@
+# Searxng
+
+[Searxng](https://github.com/searxng/searxng) is a free internet metasearch engine which aggregates results from various search services and databases. Users are neither tracked nor profiled.
+
+## Getting Started
+
+```nix
+# In `perSystem.process-compose.<name>`
+{
+  services.searxng."instance-name" = {
+    enable = true;
+    settings = {
+      use_default_settings = true;
+      server.port = 1234;
+      server.bind_address = "127.0.0.1";
+      server.secret_key = "foobar";
+    };
+  };
+}
+```
--- a/doc/searxng.md
+++ b/doc/searxng.md
@@ -9,11 +9,12 @@
 {
   services.searxng."instance-name" = {
     enable = true;
+    port = 1234;
+    host = "127.0.0.1";
+    secret_key = "my-secret-key";
     settings = {
-      use_default_settings = true;
-      server.port = 1234;
-      server.bind_address = "127.0.0.1";
-      server.secret_key = "foobar";
+      doi_resolvers."dummy" = "http://example.org";
+      default_doi_resolver = "dummy";
     };
   };
 }
--- a/nix/searxng.nix
+++ b/nix/searxng.nix
@@ -7,20 +7,6 @@
 let
   inherit (lib) types;
   yamlFormat = pkgs.formats.yaml { };
-
-  settingType =
-    with types;
-    (oneOf [
-      bool
-      int
-      float
-      str
-      (listOf settingType)
-      (attrsOf settingType)
-    ])
-    // {
-      description = "JSON value";
-    };
 in
 {
   options = {
@@ -40,7 +26,7 @@ in
     };
 
     settings = lib.mkOption {
-      types = yamlFormat.type;
+      type = yamlFormat.type;
       default = {
         server.secret_key = "secret";
         server.limiter = false;
@@ -65,14 +51,12 @@ in
         processes = {
           "${name}" = {
             environment = {
-              SEARXNG_SETTINGS_PATH = "${pkgs.writeText "settings.yml" (
-                builtins.toJSON (
-                  lib.recursiveUpdate config.settings {
-                    use_default_settings = true;
-                    server.bind_address = config.host;
-                    server.port = config.port;
-                  }
-                )
+              SEARXNG_SETTINGS_PATH = "${yamlFormat.generate "settings.yaml" (
+                lib.recursiveUpdate config.settings {
+                  use_default_settings = true;
+                  server.bind_address = config.host;
+                  server.port = config.port;
+                }
               )}";
             };
             command = lib.getExe config.package;
--- a/nix/searxng.nix
+++ b/nix/searxng.nix
@@ -38,8 +38,7 @@ in
         }
       '';
       description = ''
-        Searxng settings. These will be merged with (taking precedence over)
-        the default configuration.
+        Searxng settings
       '';
     };
 
--- a/nix/searxng.nix
+++ b/nix/searxng.nix
@@ -1,9 +1,4 @@
-{ pkgs
-, lib
-, name
-, config
-, ...
-}:
+{ pkgs, lib, name, config, ... }:
 let
   inherit (lib) types;
   yamlFormat = pkgs.formats.yaml { };
--- a/nix/searxng.nix
+++ b/nix/searxng.nix
@@ -20,6 +20,12 @@ in
       type = types.port;
     };
 
+    use_default_settings = lib.mkOption {
+      description = "Use default Searxng settings";
+      default = true;
+      type = types.bool;
+    };
+
     settings = lib.mkOption {
       type = yamlFormat.type;
       default = {
@@ -47,9 +53,9 @@ in
             environment = {
               SEARXNG_SETTINGS_PATH = "${yamlFormat.generate "settings.yaml" (
                 lib.recursiveUpdate config.settings {
-                  use_default_settings = true;
                   server.bind_address = config.host;
                   server.port = config.port;
+                  use_default_settings = config.use_default_settings;
                 }
               )}";
             };
--- a/nix/searxng.nix
+++ b/nix/searxng.nix
@@ -26,16 +26,20 @@ in
       type = types.bool;
     };
 
+    secret_key = lib.mkOption {
+      description = "Searxng secret key";
+      default = "secret";
+      example = "secret-key";
+      type = types.str;
+    };
+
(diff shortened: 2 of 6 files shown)


nix/searxng.nix
```diff
@@ -0,0 +1,94 @@
...
+      type = types.port;
+    };
+
+    settings = lib.mkOption {
+      type = types.attrsOf settingType;
+      default = {
+        use_default_settings = true;
+        server.secret_key = "secret";
+        server.limiter = false;
+      };
```
shivaraj-bh, line 47: I tried out the defaults, seems like `use_default_settings` and `server.secret_key` are essential defaults to start the server. How about we move them to where `server.{bind_address, port}` are being set? 

This way we could keep the default as `{ }`, allowing the user to override the settings without copying over the defaults. 

Also, about `server.limiter = false;`, I didn’t really see any difference if I don’t set it. From the documentation (https://docs.searxng.org/admin/settings/settings_server.html), it seems like we should let the user handle this configuration.

shivaraj-bh: > How about we move them to where server.{bind_address, port} are being set?

By doing so, the user might no longer be able to edit `server.secret_key`, for which, we could introduce a new option `secretKey`?

shivaraj-bh (changes requested):

shivaraj-bh:
I just tried it out with the following settings:
```nix
{
  services.searxng."searxng1" = {
    enable = true;
    settings = {
      use_default_settings = true;
      server.secret_key = "secret";
      search.formats = [ "html" "json" ];
    };
  };
}
```

And tried out the `json` format output with `curl`:
```sh
❯ curl -G 'http://localhost:1234/search' \
  --data-urlencode 'q=OpenAI' \
  --data-urlencode 'format=json' | jq '.results | .[:1]'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 34975  100 34975    0     0  11099      0  0:00:03  0:00:03 --:--:-- 11099
[
  {
    "url": "https://openai.com/",
    "title": "OpenAI",
    "content": "OpenAI announces GPT-4o, a new version of its powerful language model, and ChatGPT, a free service that lets you chat with it. Learn more about their research, products, and mission to create safe and beneficial artificial general intelligence.",
    "thumbnail": "",
    "engine": "google",
    "parsed_url": [
      "https",
      "openai.com",
      "/",
      "",
      "",
      ""
    ],
    "template": "default.html",
    "engines": [
      "duckduckgo",
      "google",
      "brave"
    ],
    "positions": [
      1,
      1,
      1,
      2
    ],
    "publishedDate": "2024-05-13T00:00:00",
    "score": 14.0,
    "category": "general"
  }
]
```

pretty cool! will be really helpful to provide search engine access to agents.

nix/searxng_test.nix
```diff
@@ -0,0 +1,15 @@
+{ pkgs, ... }:
+{
+  services.searxng."searxng1".enable = true;
+
+  settings.processes.test = {
+    command = pkgs.writeShellApplication {
+      runtimeInputs = [ pkgs.curl ];
+      text = ''
+        curl http://127.0.0.1:8080
```
shivaraj-bh, line 9:
```suggestion
        curl http://127.0.0.1:1234
```

nix/searxng_test.nix
```diff
@@ -0,0 +1,15 @@
...
+      '';
+      name = "searxng-test";
+    };
+    depends_on."searxng".condition = "process_healthy";
```
shivaraj-bh, line 13:
```suggestion
    depends_on."searxng1".condition = "process_healthy";
```

shivaraj-bh (changes requested):

nix/searxng.nix
```diff
@@ -0,0 +1,94 @@
...
+    };
+
+    port = lib.mkOption {
+      description = "Searxng port to listen on";
+      default = 1234;
+      type = types.port;
+    };
+
+    settings = lib.mkOption {
+      type = types.attrsOf settingType;
```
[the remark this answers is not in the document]
drupol: Done.

drupol:
> pretty cool! will be really helpful to provide search engine access to agents.

Already using it in conjunction with open-webui on my server, that's the reason why I did this PR.

nix/searxng.nix
```diff
@@ -0,0 +1,94 @@
+{ pkgs
+, lib
+, name
+, config
+, ...
+}:
```
srid, line 6: These should be in a single line, to be consistent with the formatting of the rest of the files.

nix/searxng.nix
```diff
@@ -0,0 +1,94 @@
...
+      };
+      example = lib.literalExpression ''
+        {
+          server.secret_key = "secret";
+          server.limiter = false;
+        }
+      '';
+      description = ''
+        Searxng settings. These will be merged with (taking precedence over)
+        the default configuration.
```
srid, line 56: I don't understand the merge semantics here. What is the "default configuration"? If it is referring to lines 44-47, I don't see where the merge happens.

nix/searxng.nix
drupol: I used the default `treefmt`, how about letting the tool format the file and avoid these discussions (IMHO, it's a waste of time) ?
If there's something wrong with the formatting, let's just fix the underlying tool and fix all the files at once in a follow up PR.

nix/searxng.nix
drupol: I removed the obsolete description

nix/searxng.nix
srid: `nixpkgs-fmt`, which `treefmt` uses, **does not** *reformat* one-line arguments into multi-line arguments. The point here is that formatting should be _consistent_ with the rest of the files. You should not introduce inconsistent formatting.

drupol: OK fine.
I just pushed the changes.

srid (commented):
LGTM from my side. 

Can merge after @shivaraj-bh review (and addressing of comments) with CI fully passing (currently fails on mac).

nix/searxng.nix
shivaraj-bh: @drupol Sorry, your previous defaults were right. I realised this while I was trying to get the tests working on Linux and realised that making `use_default_settings = true;` mandatory by moving it along with `server.{bind_address, port}` might be problematic. Here’s why:

I was able to get the tests pass in Linux by using the following config:

```nix
{  
  services.searxng."searxng1" = {
    enable = true;
    settings = {
      # Can't use this in test as it tries to connect to search engines during init, which is not possible in the sandbox where `nix flake check` runs.
      # use_default_settings = true;

      # Keys mandatory for starting the applications
      server.secret_key = "secret";
      doi_resolvers."dummy" = "http://example.org";
      default_doi_resolver = "dummy";
    };
  };
}
```

The tests are currently failing because of `use_default_settings = true;`, which would have the search engines pre-configured. During init, `searxng` tries to connect to these engines. As the flake checks happen in a sandbox, we will have to disable this behaviour by not using `use_default_settings = true;`.

shivaraj-bh: Could you also use the ^ config in `searxng_test.nix`, the Linux test should pass after that. You can verify this by running `just test searxng` inside the devShell.

shivaraj-bh:
> Can merge after @shivaraj-bh review (and addressing of comments) with CI fully passing (currently fails on mac).

For the time-being, we can disable tests for this package on Darwin (like we do for grafana), with a comment linking to upstream issue: https://github.com/NixOS/nixpkgs/issues/321329.

shivaraj-bh (approved):

doc/searxng.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Getting Started
+
+```nix
+# In `perSystem.process-compose.<name>`
+{
+  services.searxng."instance-name" = {
+    enable = true;
+    settings = {
+      use_default_settings = true;
```
drupol, line 13: wait

drupol:
I think i'm done.

drupol:
Top! thanks.

Issue #242, opened 2024-06-20
Document the use of a local service definition


drupol:
Hello,

I would like to have some documentation on how to have custom local services definitions that are not meant to be upstreamed in `services-flake`.

Do you have some info on how to do that?

Thanks.

shivaraj-bh:
Someone asked the same question on Zulip sometime ago: https://nixos.zulipchat.com/#narrow/stream/414011-services-flake/topic/.E2.9C.94.20custom.20local.20service/near/440612147

Here’s what the message says:

Yes, there is. See the pgweb service here: https://github.com/juspay/services-flake/blob/5ffae902f78b8f31416e5667ab95c40ee7a5ebef/example/simple/flake.nix#L42-L61

And here’s an example of a service that doesn’t exist upstream but is only specific to the project: https://github.com/nammayatri/nammayatri/blob/main/Backend/nix/services/postgres-with-replica.nix

And one more example: https://github.com/nammayatri/passetto/blob/nixify/process-compose.nix. This service is imported in nammayatri: https://github.com/nammayatri/nammayatri/blob/e8032f1fac3581b9062e2469dfc778d2913d3665/Backend/nix/services/nammayatri.nix#L32

and it is configured like: https://github.com/nammayatri/nammayatri/blob/e8032f1fac3581b9062e2469dfc778d2913d3665/Backend/nix/services/nammayatri.nix#L285-L297

shivaraj-bh:
I am also working on documenting this more formally now.

drupol:
Cool !

I tried to do this thing: https://github.com/drupol/wip-services-flake-llm

But I'm getting nowhere at the moment. I wish it could be so simple, but it is not yet.

The idea is to mimic what's done in `services-flake` and be able to import services easily.

Do you have a clue on what's wrong in there?

shivaraj-bh:
Aah, infinite recursions are painful to debug.

I took a guess and was the way `multiService` was being used, I fixed it here: https://github.com/shivaraj-bh/wip-services-flake-llm/commit/8b3016d077d68365ca0a3f76295b52b5b416f531

Although, it would be interesting to see why was there infinite recursion, because by looking at it, it seemed right.

drupol:
Thank you mate <3, I really appreciate. Now I'm going to watch at your changes and hack a bit to try to understand this correctly.

srid:
Ah, I forgot to note. Can we put this page *under* "Guide"? 

(The URL should remain thee same)

[...]

PR #243, opened 2024-06-20
docs(custom-service): Document the use of a custom service definition

shivaraj-bh:
Resolves https://github.com/juspay/services-flake/issues/242

Preview:
```sh
nix run "github:juspay/services-flake/custom-service?dir=doc"
```


doc/custom-service.md
```diff
@@ -0,0 +1,166 @@
+# Custom service
+
+While using `services-flake` you are not just limited to the [[services|services supported here]]. You can define your own service, and use it in your flake.
```
srid, line 3:
```suggestion
When using `services-flake` you are not just limited to the [[services|builtin services]]. You can also define your own service.
```

doc/custom-service.md
```diff
@@ -0,0 +1,166 @@
...
+
+By default, the services provided in `services-flake` support multiple instances, allowing you to run several instances of the same service simultaneously. However, you also have the option to create custom single-instance services. In the following sections, we’ll explore how to define custom services of both types.
```
srid, line 5:
```suggestion
By default, `services-flake` supports multiple instances for each service, allowing you to run several instances of the same service simultaneously. However, you also have the option to create custom single-instance services. In the following sections, we’ll explore how to define custom services of both types.
```

doc/custom-service.md
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+
+Finally, `nix run`:
+![[single-instance-hello.png]]
+
+{#multi-instance}
+## Multi-instance service
+
+For this purpose, `services-flake` exports a [multiService](https://github.com/juspay/services-flake/blob/647bff2c44b42529461f60a7fe07851ff93fb600/nix/lib.nix#L1-L34) library function. The aim is to provide an interface wherein the user just writes the configuration like they would for a single instance service, and the library takes care of creating multiple instances of the service.
```
srid, line 79:
```suggestion
For this purpose, `services-flake` exports a [multiService](https://github.com/juspay/services-flake/blob/647bff2c44b42529461f60a7fe07851ff93fb600/nix/lib.nix#L1-L34) library function. It aims to provide an interface wherein the user just writes the configuration like they would for a single instance service, and the library takes care of creating multiple instances of the service.
```

srid (approved):
Looks great.

(Ignore the Linux CI failure; I'm fixing it...)

PR #258, opened 2024-07-04
feat: Allow overriding namespace + make default namespace fully qualified

srid:
**PR description**

- The user can now set `namespace` for the processes under each service
- Default namespace is now `${service}.${name}` (previously, just `${name}`)

**Tasks**

- [x] Do this for postgres, to begin with.
- [x] Rest of the services

image

After the user overrides it:

image

After the user overrides the local processes as well:

image


srid:
The other thing we can do here is to support a special service name, `"default"`. Handy for cases where you run exactly 1 service, so both default namespace and the process name can drop it. Instead of `db-init`, it would just be `postgres-init`.

srid:
To be squash merged.

nix/lib.nix
```diff
@@ -8,25 +8,62 @@
...
+      service = lib.pipe mod [
         builtins.baseNameOf
         (lib.strings.splitString ".")
         builtins.head
       ];
+      serviceModule = { config, name, ... }: {
+        options = {
+          namespace = lib.mkOption {
+            description = ''
+              Namespace for the ${service} service
```
shivaraj-bh, line 20:
```suggestion
              Namespace for the ${service}.${name} service
```

nix/lib.nix
```diff
@@ -8,25 +8,62 @@
...
+            '';
+            default = "${service}.${name}";
+            type = lib.types.str;
+          };
+          outputs.defaultProcessSettings = lib.mkOption {
+            type = lib.types.deferredModule;
+            internal = true;
+            readOnly = true;
+            description = ''
+              Default settings for all processes under the ${service} service
```
shivaraj-bh, line 30:
```suggestion
              Default settings for all processes under the ${service}.${name} service
```

nix/postgres/default.nix
```diff
@@ -301,7 +300,7 @@ in
               in
               {
                 command = setupScript;
-                namespace = name;
+                inherit (config) namespace;
```
shivaraj-bh, line 303:
```suggestion
```

nix/postgres/default.nix
```diff
@@ -329,6 +328,7 @@ in
                 ] ++ (lib.optional (config.superuser != null) "-U ${config.superuser}");
               in
               {
+                inherit (config) namespace;
```
shivaraj-bh, line 331:
```suggestion
```

shivaraj-bh (changes requested):

nix/lib.nix
srid: If `name` is dynamic, how will the docs render?

nix/lib.nix
srid: (See below)

nix/lib.nix
shivaraj-bh: My bad, I missed that.

PR #261, opened 2024-07-05
refactor: Do not repeat `dataDir`

### LINKED ISSUE #200: Refactor(DRY): `./data/${name}`
Easy to break this convention in new PRs

image

e.g.:  https://github.com/juspay/services-flake/pull/195#discussion_r1615217405

shivaraj-bh:
resolves #200

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -15,6 +15,7 @@
       ];
       serviceModule = { config, name, ... }: {
         options = {
+          enable = lib.mkEnableOption "Enable the ${service}.<name> service";
           dataDir = lib.mkOption {
             type = lib.types.str;
             default = "./data/${name}";
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -15,6 +15,11 @@
       ];
       serviceModule = { config, name, ... }: {
         options = {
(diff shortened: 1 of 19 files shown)


nix/lib.nix
```diff
@@ -15,6 +15,11 @@
       ];
       serviceModule = { config, name, ... }: {
         options = {
+          dataDir = lib.mkOption {
+            type = lib.types.str;
+            default = "./data/${name}";
+            description = "The ${service}.<name> data directory";
```
shivaraj-bh, line 21: @srid How does this look for a description? In the docs the option will be rendered as `<name>` anyways?

context: https://github.com/juspay/services-flake/pull/258#discussion_r1666999402

shivaraj-bh:
I also think we can refactor `enable` option in the similar fashion

nix/lib.nix
srid: Sure (but hard to say definitely without rendering the docs though)

srid, line 21:
```suggestion
            description = "The directory where all data for `${service}.<name>` is stored";
```

srid (approved):

PR #311, opened 2024-09-04
feat: Add `max_restarts` to all services to avoid restarting indefinitely

### LINKED ISSUE #309: postgres: Don't infinitely keep restarting when DB couldn't start
This just goes on an infinite loop trying to restart when it will never succeed. Should we not crash the whole process?

image

cf. https://github.com/juspay/services-flake/blob/113f8cbe962720d7d1d76d04700f3d441dada9f6/nix/services/postgres/default.nix#L337

shivaraj-bh:
resolves #309

### DIFF
--- a/nix/services/apache-kafka.nix
+++ b/nix/services/apache-kafka.nix
@@ -177,7 +177,10 @@ with lib;
                 failure_threshold = 5;
               };
 
-              availability.restart = "on_failure";
+              availability = {
+                restart = "on_failure";
+                max_restarts = 5;
+              };
             };
         };
       };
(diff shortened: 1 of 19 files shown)


srid (commented):
Wasn't this used to be the case before? I vaguely remember process restarts never being infinite to begin with (I've seen processes enter into 'exited' state in process-compose window).

In any case, this change only affects services. Should we make it a default in process-compose-flake? Actually why not make it a default in `process-compose` itself?

Incidentally, there's no guarantee that a new service added to the repo will follow the same max_restarts convention. It is easy to miss that, so we should think about solving this more fundamentally.

PR #313, opened 2024-09-07
Fixed links in usage examples

secobarbital:
nix/*_test.nix -> nix/services/*_test.nix


shivaraj-bh:
I should’ve used permalinks, but any changes in the API would mean updating the permalink, so this should be fine.

Thanks for noticing and fixing it.

PR #314, opened 2024-09-07
Added memcached service

secobarbital:


nix/services/memcached_test.nix
```diff
@@ -0,0 +1,23 @@
...
+    in
+    {
+      command = pkgs.writeShellApplication {
+        runtimeInputs = [
+          cfg.package
+          pkgs.gnugrep
+          pkgs.netcat
+        ];
+        text = ''
+          echo -e "stats\nquit" | nc 127.0.0.1 11211 | grep -q "STAT version"
```
shivaraj-bh, line 17:
```suggestion
          echo -e "stats\nquit" | nc 127.0.0.1 11211 | grep "STAT version"
```
For more visibility in the logs:
```sh
memcached-test> + process-compose --no-server -t=false
memcached-test> [test   ] STAT version 1.6.27
memcached-test> + set +x
```

nix/services/memcached.nix
```diff
@@ -0,0 +1,67 @@
+# Based on https://github.com/cachix/devenv/blob/main/src/modules/services/memcached.nix
+{ pkgs
+, lib
+, name
+, config
+, ...
+}:
```
srid, line 7: Could you put this in single line to be consistent with the formatting of the rest of the project?

test/flake.nix
srid, line 1: Can you revert all formatting changes to this file? PR diffs should be minimal.

doc/memcached.md
```diff
@@ -0,0 +1,6 @@
+# Memcached
+
+>[!warning] 🚧 WIP
```
srid, line 3: Is memcached service in fact a work in progress? If not, remove this - and add some basic docs: at minimum an introduction (as to what memcached is).

#316

nix/services/memcached.nix
secobarbital: Will do. I remember seeing treefmt formatting it in a pre-commit-hook, must have been dreaming.

srid: It could have been your editor doing an auto formatting on save.

secobarbital: Yes, my editor definitely did that, but I saw treefmt changing 4 files. But now when I manually call it, it doesn't do anything. Will have to manually revert those changes.
```
% just fmt
treefmt
0 files changed in 5ms (found 109, matched 54, cache misses 0)
```

shivaraj-bh:
Thanks for the contribution!

PR #331, opened 2024-09-17
elasticsearch: copy plugins to `dataDir`

### LINKED ISSUE #325: Failed to load elasticsearch plugin
I am able to run elasticsearch service just fine on its own, but when I add a plugin, it fails with the following error:

```
uncaught exception in thread [main]
java.lang.IllegalStateException: failed to load plugin analysis-icu while checking for jar hell
Likely root cause: java.security.AccessControlException: access denied ("java.io.FilePermission" "/nix/store/b3n8yyfsvzf1ms2sf7f8qcmx9zy3fslz-elasticsearch-analysis-icu-7.17.16/plugins/analysis-icu/analysis-icu-client-7.17.16.jar" "read")
	at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
	at java.base/java.security.AccessController.checkPermission(AccessController.java:897)
	at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:322)
	at java.base/java.lang.SecurityManager.checkRead(SecurityManager.java:661)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:260)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:200)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:348)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:319)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:258)
	at org.elasticsearch.jdk.JarHell.checkJarHell(JarHell.java:180)
	at org.elasticsearch.plugins.PluginsService.checkBundleJarHell(PluginsService.java:689)
	at org.elasticsearch.plugins.PluginsService.loadBundles(PluginsService.java:531)
	at org.elasticsearch.plugins.PluginsService.<init>(PluginsService.java:170)
	at org.elasticsearch.node.Node.<init>(Node.java:411)
	at org.elasticsearch.node.Node.<init>(Node.java:309)
	at org.elasticsearch.bootstrap.Bootstrap$5.<init>(Bootstrap.java:234)
	at org.elasticsearch.bootstrap.Bootstrap.setup(Bootstrap.java:234)
	at org.elasticsearch.bootstrap.Bootstrap.init(Bootstrap.java:434)
	at org.elasticsearch.bootstrap.Elasticsearch.init(Elasticsearch.java:169)
	at org.elasticsearch.bootstrap.Elasticsearch.execute(Elasticsearch.java:160)
	at org.elasticsearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:77)
	at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:112)
	at org.elasticsearch.cli.Command.main(Command.java:77)
	at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:125)
	at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:80)
```

I verified that the file in question is indeed there, but I read that elasticsearch [might not like accessing plugins via symlinks](https://discuss.elastic.co/t/after-enable-elasticsearch-plugin-sh-the-es-cannot-start/362501/5), which seems to be [how we are doing it](https://github.com/juspay/services-flake/blob/main/nix/services/elasticsearch.nix#L150-L155).

Here's my services declaration:

```

              services.elasticsearch."es1" = {
                inherit imports;
                enable = true;
                plugins = [
                  pkgs.elasticsearchPlugins.analysis-icu
                  pkgs.elasticsearchPlugins.analysis-kuromoji
                ];
              };
```

shivaraj-bh:
resolves #325

### DIFF
--- a/nix/services/elasticsearch.nix
+++ b/nix/services/elasticsearch.nix
@@ -147,8 +147,11 @@ in
                 export ES_HOME ES_JAVA_OPTS ES_PATH_CONF
 
                 # Install plugins
-                rm -f "${config.dataDir}/plugins"
-                ln -sf ${esPlugins}/plugins "${config.dataDir}/plugins"
+                rm -rf "${config.dataDir}/plugins"
+                cp -rL ${esPlugins}/plugins "${config.dataDir}/plugins"
+                find "${config.dataDir}/plugins" -type f -exec chmod 0700 {} \;
+                find "${config.dataDir}/plugins" -type d -exec chmod 0700 {} \;
+
                 rm -f "${config.dataDir}/lib"
                 ln -sf ${config.package}/lib "${config.dataDir}/lib"
                 rm -f "${config.dataDir}/modules"
--- a/nix/services/elasticsearch.nix
+++ b/nix/services/elasticsearch.nix
@@ -149,7 +149,6 @@ in
                 # Install plugins
                 rm -rf "${config.dataDir}/plugins"
                 cp -rL ${esPlugins}/plugins "${config.dataDir}/plugins"
-                find "${config.dataDir}/plugins" -type f -exec chmod 0700 {} \;
                 find "${config.dataDir}/plugins" -type d -exec chmod 0700 {} \;
 
                 rm -f "${config.dataDir}/lib"
--- a/nix/services/elasticsearch.nix
+++ b/nix/services/elasticsearch.nix
@@ -149,7 +149,7 @@ in
                 # Install plugins
                 rm -rf "${config.dataDir}/plugins"
                 cp -rL ${esPlugins}/plugins "${config.dataDir}/plugins"
-                find "${config.dataDir}/plugins" -type d -exec chmod 0700 {} \;
+                find "${config.dataDir}/plugins" -type d -exec chmod u+x {} \;
 
                 rm -f "${config.dataDir}/lib"
                 ln -sf ${config.package}/lib "${config.dataDir}/lib"
--- a/nix/services/elasticsearch_test.nix
+++ b/nix/services/elasticsearch_test.nix
@@ -1,11 +1,19 @@
 { pkgs, config, ... }: {
-  services.elasticsearch."es1".enable = true;
+  services.elasticsearch."es1" = {
+    enable = true;
+    plugins = [
+      pkgs.elasticsearchPlugins.analysis-icu
+    ];
+  };
   settings.processes.test =
     {
       command = pkgs.writeShellApplication {
         runtimeInputs = [ pkgs.curl ];
         text = ''
           curl 127.0.0.1:9200/_cat/health 
+
+          echo "Verify plugin installation"
+          curl 127.0.0.1:9200/_cat/plugins | grep 'analysis-icu'
         '';
         name = "elasticsearch-test";
       };


shivaraj-bh:
@secobarbital can you check if this works for you?

Also, I have included a test to check for plugin installation.

secobarbital:
Wow, thank you for the quick turnaround! Yes, this does work for me, it is almost the same as the workaround I came back to report:
```
                cp -RL ${esPlugins}/plugins "${config.dataDir}"
                find "${config.dataDir}/plugins" -type d -exec chmod u+w {} \;
```
I find that only the directories needed their permissions fixed to get it to work, not the files.

However, I also found this in [the official elasticsearch Dockerfile](https://github.com/elastic/dockerfiles/blob/5e07485ccc161d0661f8a04462b6cc6f565e2573/elasticsearch/Dockerfile#L67-L71) and I wonder if we should follow their permissions as canonical. It looks like they intentionally package the files with no permissions and expect consumer to modify them?
```
RUN sed -i -e 's/ES_DISTRIBUTION_TYPE=tar/ES_DISTRIBUTION_TYPE=docker/' bin/elasticsearch-env && \
    mkdir data && \
    mv config/log4j2.properties config/log4j2.file.properties && \
    mv config/log4j2.docker.properties config/log4j2.properties && \
    find . -type d -exec chmod 0555 {} + && \
    find . -type f -exec chmod 0444 {} + && \
    chmod 0555 bin/* jdk/bin/* jdk/lib/jspawnhelper modules/x-pack-ml/platform/linux-*/bin/* && \
    chmod 0775 bin config config/jvm.options.d data logs plugins && \
    find config -type f -exec chmod 0664 {} +
```

nix/services/elasticsearch.nix
```diff
@@ -147,8 +147,11 @@ in
                 export ES_HOME ES_JAVA_OPTS ES_PATH_CONF
 
                 # Install plugins
-                rm -f "${config.dataDir}/plugins"
-                ln -sf ${esPlugins}/plugins "${config.dataDir}/plugins"
+                rm -rf "${config.dataDir}/plugins"
+                cp -rL ${esPlugins}/plugins "${config.dataDir}/plugins"
+                find "${config.dataDir}/plugins" -type f -exec chmod 0700 {} \;
```
secobarbital, line 152: We don't really need to change the perms of the regular files, and I don't think we should make them executable.
```suggestion
```

nix/services/elasticsearch.nix
```diff
@@ -147,8 +147,11 @@ in
...
+                find "${config.dataDir}/plugins" -type d -exec chmod 0700 {} \;
```
secobarbital, line 153: I think we don't need to remove permissions from the group and other. The original perms on the directories are 555.
```suggestion
                find "${config.dataDir}/plugins" -type d -exec chmod u+x {} \;
```

secobarbital (commented):
This works! Coupa of slight tweaks to suggest.

shivaraj-bh:
Thanks @secobarbital

PR #332, opened 2024-09-21
fix(elasticsearch): Fixed permission to add to directories in elasticsearch plugins

secobarbital:
We need to add the write permission to directories in elasticsearch plugins, not execute, which they already have.

Sorry @shivaraj-bh this was a typo in my suggestion in #331

### DIFF
--- a/nix/services/elasticsearch.nix
+++ b/nix/services/elasticsearch.nix
@@ -149,7 +149,7 @@ in
                 # Install plugins
                 rm -rf "${config.dataDir}/plugins"
                 cp -rL ${esPlugins}/plugins "${config.dataDir}/plugins"
-                find "${config.dataDir}/plugins" -type d -exec chmod u+x {} \;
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
Does the test need some fixing as well? I just assumed it was all good because the test passed.

secobarbital:
Hmm… that makes sense. The problem arises on the next time you start it, when the startup script tries to delete the plugins directory, so the test would have to start it twice.

shivaraj-bh:
> The problem arises on the next time you start it, when the startup script tries to delete the plugins directory, so the test would have to start it twice.

This will need some discussion in the upstream `process-compose-flake`. 

For now, I have tested this PR, we can go ahead with the merge.

shivaraj-bh:
> This will need some discussion in the upstream process-compose-flake.

https://github.com/Platonic-Systems/process-compose-flake/discussions/78

secobarbital:
Thank you!

PR #338, opened 2024-09-23
Check for conventional commits in CI

shivaraj-bh:


shivaraj-bh:
https://github.com/juspay/services-flake/actions/runs/11004484171/job/30555489649?pr=338

image


Tested for failure, force pushing now to pass CI

dev/flake.nix
```diff
@@ -32,6 +33,31 @@
...
+              # rev range using the `default_branch` branch name (e.g. main..HEAD), doesn't work in Github Actions
+              latest_default_commit=$(git rev-parse origin/"$default_branch")
+
+              current_commit=$(git rev-parse HEAD)
+
+              if [ "$latest_default_commit" = "$current_commit" ]; then
+                  echo "No commits to check between $default_branch and HEAD."
+              else
+                  cz check --rev-range "$latest_default_commit"..HEAD
+              fi
```
shivaraj-bh, line 57: This implementation isn’t generic enough, i.e, it always assumes that the current branch is going to be merged to `main`, which is okay for now. We can re-think this later if we have a complicated branch tree.

dev/flake.nix
```diff
@@ -32,6 +33,32 @@
             nixpkgs-fmt.enable = true;
           };
         };
+
+        # TODO: Move this under `CIApps` once `omnix` supports running them automatically
```
srid, line 37: What is `CIApps`?

dev/flake.nix
srid: `$GITHUB_BASE_REF`? 

https://stackoverflow.com/a/65656616/55246

srid (approved):

dev/flake.nix
shivaraj-bh: Custom flake-schema that can be detected by omnix to run everything under it as custom app steps.

Issue #158, opened 2024-03-21
Option docs


shivaraj-bh:
Provide searchable option docs for each service.

srid:
We add services-flake to https://flake.parts/ (which renders the options doc)

> Provide searchable option docs for each service.

Eventually, I'd like nix-browser to do this as well.

danielo515:
Will this include all options you can pass to services? I had to read the code to see that I can configure postgres by providing a settings attrset

PR #339, opened 2024-09-26
feat: Add support for mongodb

greg-hellings:
Adds support for running MongoDB, based off the Redis workflow.

### DIFF
--- a/doc/mongodb.md
+++ b/doc/mongodb.md
@@ -1,0 +1,11 @@
+# MongoDB
+
+[MongoDB](https://www.mongodb.com/) is a popular document database that is available. It comes in a variety of flavors, but the version packaged by NixPkgs is generally the community edition.
+
+Because of the licensing of MongoDB, the Nixpkgs derivation that provides the binaries of this database are generally not built and cached by the public nixpkgs cache. Because of this, the initial launch time can be very slow, the first time you use this service locally. On a laptop this time ran to about 3 hours for the initial compile. After the initial build, the start up should be very fast, as with other services, provided you do not update your flake.lock.
+
+If you are using this for your own development, you should either put in place a [local binary cache](https://nixos.wiki/wiki/Binary_Cache) to match your flake, or be aware that the first time you spin up the service it could possibly take a long time to build.
+
+## Usage example
+
+<https://github.com/juspay/services-flake/blob/main/nix/services/mongodb_test.nix>
--- a/nix/services/mongodb.nix
+++ b/nix/services/mongodb.nix
@@ -1,0 +1,133 @@
+{ pkgs, lib, name, config, ... }:
+let
+  inherit (lib) types;
+in
+{
+  options = {
+    package = lib.mkPackageOption pkgs "mongodb" { };
+
+    bind = lib.mkOption {
+      type = types.nullOr types.str;
+      default = "127.0.0.1";
+      description = ''
+        The IP interface to bind to.
+        `null` means "all interfaces".
+      '';
+      example = "127.0.0.1";
+    };
+
+    port = lib.mkOption {
+      type = types.port;
+      default = 27017;
+      description = ''
+        The TCP port to accept connections.
+      '';
+    };
+
+    user = lib.mkOption {
+      type = types.str;
+      default = "default";
+      description = ''
+        The name of the first user to create in
+        Mongo.
+      '';
+      example = "my_user";
+    };
+
+    password = lib.mkOption {
+      type = types.str;
+      default = "password";
+      description = ''
+        The password of the user to configure
+        for initial access.
+      '';
+    };
+
+    extraConfig = lib.mkOption {
+      type = types.lines;
+      default = "";
+      description = "Additional text to be appended to `mongodb.conf`.";
+    };
+  };
+
+  config = {
+    outputs = {
+      settings = {
+        processes = {
+          "${name}" =
+            let
+              mongoConfig = pkgs.writeText "mongodb.conf" ''
+                net.port: ${toString config.port}
+                net.bindIp: ${config.bind}
+                storage.dbPath: ${config.dataDir}
+                ${config.extraConfig}
+              '';
+
+              startScript = pkgs.writeShellApplication {
+                name = "start-mongodb";
+                runtimeInputs = [ pkgs.coreutils config.package ];
+                text = ''
+                  export MONGODATA="${config.dataDir}"
+
+                  if [[ ! -d "$MONGODATA" ]]; then
+                    mkdir -p "$MONGODATA"
+                  fi
+
+                  exec mongod --config "${mongoConfig}"
+                '';
+              };
+            in
+            {
+              command = startScript;
+
+              readiness_probe = {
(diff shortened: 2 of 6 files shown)


srid:
Thanks! Could you add some basic docs? See

https://community.flake.parts/services-flake/contributing#service-doc

shivaraj-bh:
image

This dependency seems to be broken on Darwin. If the tests run on Linux you can conditionally disable it on Darwin (linking to upstream issue, if any):

https://github.com/juspay/services-flake/blob/23e40bdae913b64ffaeb774201b5a6f868ce19ca/test/flake.nix#L50-L55

adrian-gierakowski:
You can use python311 since the default now is 3.12 which suffers from the breakage

adrian-gierakowski:
Btw mongodb is never built by official nix hydra due to licence but I’ve seen it in devenv’s cache (including macOS). Might have to pull it from older nixpkgs commit

greg-hellings:
I'll get on docs right now (sorry, I was at a company off-site last week when I put the PR together and haven't gotten back to it since getting home), along with setting the restriction on Darwin if it is indeed completely broken there.

As for there not being a binary cache, that was apparent the first time I fired up the test environment while I was rolling this for some dev work I was doing. On my work issued laptop (recent gen mobile Intel i7) the compile takes around 2 to 3 hours the first time it compiles Mongo, but it did go through without a hitch. After that first build, it's just as fast as anything else that you have locally cached. I know there are ways to set binary cache hints in a flake or such. But that would necessitate integrating with a binary cache, so I don't know if that's what you want to do here. Maybe I will just put a disclaimer in the docs letting people know that compiling.

From the point of view of the CI testing in this repo, though, I don't know what you might want to do.

greg-hellings:
I've updated the test to be dependent on the `jaraco-path.meta.broken` value. If that gets unblocked in the future by fixing jaraco-path, this test will be automatically re-enabled. If the dependency gets worked around in the future then this can be revisted.

doc/mongodb.md
```diff
@@ -0,0 +1,22 @@
...
+## Options
+
+This module uses most of the same common options that other database modules in this flake use. Some of the most sommon ones are
+
+* `bind` - the address to bind to. Default: 127.0.0.1
+* `port` - the port to serve on. Default: 27017
+* `user` - the MongoDB user to create. Default: default
+* `password` - the password for the MongoDB user. Default: password
+* `extraConfig` - arbitrary extra lines to add to the MongoDB config file. Default: empty string
+* `dataDir` - directory to store data files in. See [datadir](https://github.com/juspay/services-flake/blob/main/doc/datadir.md)
```
shivaraj-bh, line 18: For options, we are planning to generate them in https://github.com/juspay/services-flake/issues/158. You can remove them from here.

greg-hellings: Ah, that explains why I haven't been able to find them without trolling through the Nix source.

test/flake.nix
```diff
@@ -51,6 +51,10 @@
             "${inputs.services-flake}/nix/services/tika_test.nix"
             "${inputs.services-flake}/nix/services/weaviate_test.nix"
             "${inputs.services-flake}/nix/services/zookeeper_test.nix"
+          ] ++ lib.optionals (!pkgs.python3.pkgs.jaraco-path.meta.broken) [
```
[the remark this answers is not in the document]
greg-hellings: Mmm, I wish that the Mongo package itself had a better value for its `broken` tag, because I would rather key off of that. In the absence of it, I have moved to just checking for `isLinux`.

nix/services/mongodb.nix
```diff
@@ -0,0 +1,131 @@
...
+                net.bindIp: ${config.bind}
+                storage.dbPath: ${config.dataDir}
+                ${config.extraConfig}
+              '';
+
+              startScript = pkgs.writeShellApplication {
+                name = "start-mongodb";
+                runtimeInputs = [ pkgs.coreutils config.package ];
+                text = ''
+                  set -euo pipefail
```
shivaraj-bh, line 70: This line can be removed as `writeShellApplication` sets it by default: https://github.com/NixOS/nixpkgs/blob/0b86640b433332928f978942427300733ebca24c/pkgs/build-support/trivial-builders/default.nix#L253C23-L253C53

nix/services/mongodb.nix
```diff
@@ -0,0 +1,131 @@
...
+                restart = "on_failure";
+                max_restarts = 5;
+              };
+            };
+          "${name}-configure" =
+            let
+              configScript = pkgs.writeShellApplication {
+                name = "configure-mongo";
+                text = ''
+                  if ! test -e "${config.dataDir}/.auth_configured"; then
```
shivaraj-bh, line 105: An else condition `echo`ing that the database is already configured will be helpful?

shivaraj-bh (approved):
LGTM! Just a few more changes and I can merge

nix/services/mongodb.nix
shivaraj-bh: > This line can be removed

I am referring to `set -euo pipefail`

greg-hellings: Removed.

nix/services/mongodb.nix
greg-hellings: Good idea. I've added one that should prompt the user how to remove the configuration file as well. This setup does not actually run Mongo in a mode that enforces access control - that is a more advanced feature. But in my testing, at least, it configures the optional user and password that the user specifies. The user can still connect to it without authing as well.

shivaraj-bh:
Thanks!

drupol:
Nice stuff ! Just a very quick information, [`mongodb-ce`](https://search.nixos.org/packages?channel=unstable&show=mongodb-ce&from=0&size=50&sort=relevance&type=packages&query=mongodb-ce) can be used to use pre-compiled binaries.

adrian-gierakowski:
> Nice stuff ! Just a very quick information, [`mongodb-ce`](https://search.nixos.org/packages?channel=unstable&show=mongodb-ce&from=0&size=50&sort=relevance&type=packages&query=mongodb-ce) can be used to use pre-compiled binaries.

What do you mean? They it will be built official nix hydra and available nix official binary cache?

Doesn’t looks like this to me: licence is the same as original mongodb package and there is not completed builds on hydra: https://hydra.nixos.org/job/nixos/trunk-combined/nixpkgs.mongodb-ce.x86_64-linux/latest

drupol:
This is the pre compiled binories from mongodb directly.

shivaraj-bh:
> This is the pre compiled binories from mongodb directly.

Damn! I tried replacing `mongodb` package with `mongodb-ce` and the tests pass (see: https://github.com/juspay/services-flake/pull/360).

@drupol Thanks for packaging this upstream!

shivaraj-bh:
I will also document this information in https://github.com/juspay/services-flake/pull/360, just in case someone doesn’t want to use pre-built binaries.

greg-hellings:
Oh dang, that's sweet. Especially because my laptop sometimes fails to build mongodb due to the OOM-killer terminating some of the link steps!

adrian-gierakowski:
> This is the pre compiled binories from mongodb directly.

Ah, cool. Should have looked more closely

PR #340, opened 2024-09-27
chore(pg): find util used but it is not default in nixos docker images

dzmitry-lahoda:


shivaraj-bh:
Thanks! 

For completeness:
`find` is used in: https://github.com/juspay/services-flake/blob/67f91ab4071f4ab4044fd44a58a7e09fdf8ad1d2/nix/services/postgres/setup-script.nix#L14 but the package is not included in the `runtimeInputs` of the setup script: https://github.com/juspay/services-flake/blob/67f91ab4071f4ab4044fd44a58a7e09fdf8ad1d2/nix/services/postgres/setup-script.nix#L79

shivaraj-bh:
@dzmitry-lahoda I didn’t quite get the title of the PR, to be specific "not default in nixos docker images”, could you elaborate?

dzmitry-lahoda:
yeah, i meant nor nixpkgs docker tools nor https://github.com/nlewo/nix2container images by default include find (nor curl too btw).

PR #348, opened 2024-10-04
refactor: consolidate all settings in `defaultProcessSettings`

### LINKED ISSUE #317: Re-evaluate `max_restarts`
Should we make it a default in process-compose-flake? Or perhaps file an issue in process-compose itself asking what the recommend default should be, and why that default is not in process-compose itself.

>  Wasn't this used to be the case before? I vaguely remember process restarts never being infinite to begin with (I've seen processes enter into 'exited' state in process-compose window).
> 
> In any case, this change only affects services. Should we make it a default in process-compose-flake? Actually why not make it a default in `process-compose` itself?
> 
> Incidentally, there's no guarantee that a new service added to the repo will follow the same max_restarts convention. It is easy to miss that, so we should think about solving this more fundamentally.

_Originally posted by @srid in https://github.com/juspay/services-flake/pull/311#pullrequestreview-2280223552_

shivaraj-bh:
resolves #317

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -38,6 +38,18 @@
               '';
               default = {
                 namespace = lib.mkDefault config.namespace;
+                # https://github.com/F1bonacc1/process-compose#-auto-restart-if-not-healthy
+                availability = {
+                  restart = lib.mkDefault "on_failure";
+                  max_restarts = lib.mkDefault 5;
+                };
+                readiness_probe = {
+                  initial_delay_seconds = lib.mkDefault 2;
+                  period_seconds = lib.mkDefault 10;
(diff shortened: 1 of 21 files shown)


srid (approved):

shivaraj-bh:
This PR will have to be reverted because it falsely assumes that even processes other than the one named `name` (`name` from `services.<service>.<name>`) also takes the same defaults, which is not true.

For example, when `<name>-init` process of `postgres` fails, the setting will restart the process, even though that is not the desired behaviour.

This wasn’t caught in CI because we don’t test failure of `init` process of any service, guess I will add them in the PR that reverts this change.

PR #353, opened 2024-10-09
feat(redis): Support using Unix socket

### LINKED ISSUE #352: Support using Redis with UDS
Redis has the option to connect via unix domain sockets. I have such a setup running with
```nix
services.redis."r1" = {
    enable = true;
    port = 0; # disable listening via TCP
    extraConfig = ''
      locale-collate "C"
      unixsocket ../redis.socket
      unixsocketperm 777
    '';
  };
};
settings.processes."r1".readiness_probe.exec.command = lib.mkForce "true";
```
The `../redis.socket` places the socket in the data directory instead of next to the `dump.rdb` file. The `locale-collate` setting should be unneeded for this, but I haven't checked if it is relevant for UDS.

The change in the readiness probe is needed because the current command, `${config.package}/bin/redis-cli -p ${toString config.port} ping` relies on TCP being active. If setup properly, this should probably perform the same ping, but through the socket.

It would be nice if there would be an option like `uds = "data/redis.socket"` to take care of all of this. Please let me know what you think :)

shivaraj-bh:
resolves #352

### DIFF
--- a/nix/services/redis.nix
+++ b/nix/services/redis.nix
@@ -25,6 +25,28 @@ in
 
         If port is set to `0`, redis will not listen on a TCP socket.
       '';
+      apply = v:
+        lib.warnIf ((config.unixSocket != null) && (v != 0)) ''
+          `${name}` is listening on both the TCP port and Unix socket, set `port = 0;` to listen on only the Unix socket
+        ''
+          v;
+    };
+
+    unixSocket = lib.mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        The path to the socket to bind to.
+
+        If a relative path is used, it will be relative to `dataDir`.
+      '';
+    };
+
+    unixSocketPerm = lib.mkOption {
+      type = types.int;
+      default = 660;
+      description = "Change permissions for the socket";
+      example = 600;
     };
 
     extraConfig = lib.mkOption {
@@ -43,6 +65,8 @@ in
               redisConfig = pkgs.writeText "redis.conf" ''
                 port ${toString config.port}
                 ${lib.optionalString (config.bind != null) "bind ${config.bind}"}
+                ${lib.optionalString (config.unixSocket != null) "unixsocket ${config.unixSocket}"}
+                ${lib.optionalString (config.unixSocket != null) "unixsocketperm ${builtins.toString config.unixSocketPerm}"}
                 ${config.extraConfig}
               '';
 
@@ -65,9 +89,22 @@ in
             {
               command = startScript;
 
-              readiness_probe = {
-                exec.command = "${config.package}/bin/redis-cli -p ${toString config.port} ping";
-              };
+              readiness_probe =
+                let
+                  # Transform `unixSocket` by prefixing `config.dataDir` if a relative path is used
+                  transformedSocketPath =
+                    if (config.unixSocket != null && (lib.hasPrefix "./" config.unixSocket)) then
+                      "${config.dataDir}/${config.unixSocket}"
+                    else
+                      config.unixSocket;
+                in
+                {
+                  exec.command =
+                    if (transformedSocketPath != null && config.port == 0) then
+                      "${config.package}/bin/redis-cli -s ${transformedSocketPath} ${toString config.port} ping"
+                    else
+                      "${config.package}/bin/redis-cli -p ${toString config.port} ping";
+                };
             };
         };
       };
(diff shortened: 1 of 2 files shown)


nix/services/redis_test.nix
```diff
@@ -1,6 +1,12 @@
 { pkgs, config, ... }: {
   services.redis."redis1".enable = true;
 
+  services.redis."redis2" = { config, ... }: {
+    enable = true;
+    port = 0;
+    unixSocket = "./redis.sock";
```
srid, line 7: What's the significance of `port = 0`? What happens if the user specifies both `port` and `unixSocket`?

shivaraj-bh: `port = 0` disables listening on the TCP port. Setting both is also allowed by redis, in which case it listens on both `unixSocket` and TCP `port`. Hence I just display a warning if both are set instead of error it out (https://github.com/juspay/services-flake/blob/821dbf382a5ff312aec99e4ded9da46a4eee21c3/nix/services/redis.nix#L28-L32).

srid: This should be documented then: #358

shivaraj-bh: It is documented in the option description. I will also add it to redis.md

nix/services/redis.nix
```diff
@@ -65,9 +89,22 @@ in
...
               command = startScript;
 
-              readiness_probe = {
-                exec.command = "${config.package}/bin/redis-cli -p ${toString config.port} ping";
-              };
+              readiness_probe =
+                let
+                  # Transform `unixSocket` by prefixing `config.dataDir` if a relative path is used
+                  transformedSocketPath =
+                    if (config.unixSocket != null && (lib.hasPrefix "./" config.unixSocket)) then
```
niklasmohrin, line 96: I think this doesn't work as intended, if I specify `unixSocket = "redis.socket";`, which is also a relative path, the socket will be created at `data/r1/redis.socket`. I think the check should be `! lib.hasPrefix "/" config.unixSocket`.

niklasmohrin (commented):
Thanks!

I wonder if it would be worth it to unify the option names around sockets. Right now, process-compose-flake has `httpServer.uds = "path-relative-to-project-root"`, redis has `unixSocket = "path-relative-to-data-dir"` and postgres has `socketDir = "path-relative-to-project-root`" (with no way to specify the name, but just the directory; I think it's okay though). One annoyance with the process-compose uds option is that using `"data/process-compose.socket"` won't automatically create the `data/` directory and instead fail.

nix/services/redis.nix
shivaraj-bh: https://github.com/juspay/services-flake/issues/365

shivaraj-bh:
> Thanks!
> 
> I wonder if it would be worth it to unify the option names around sockets. Right now, process-compose-flake has `httpServer.uds = "path-relative-to-project-root"`, redis has `unixSocket = "path-relative-to-data-dir"` and postgres has `socketDir = "path-relative-to-project-root`" (with no way to specify the name, but just the directory; I think it's okay though). One annoyance with the process-compose uds option is that using `"data/process-compose.socket"` won't automatically create the `data/` directory and instead fail.

Usually we try to match what’s upstream so users coming from there don’t have a hard time figuring out options. For `redis` it is the same as in nixos module in `nixpkgs`. `nixpkgs` doesn’t define an option to configure socket dir for `postgres`, so we can consider renaming that for uniformity. Even `uds` in `process-compose-flake` can be re-considered.

niklasmohrin:
> Usually we try to match what’s upstream so users coming from there don’t have a hard time figuring out options. For `redis` it is the same as in nixos module in `nixpkgs`

Ah, yeah, that makes perfect sense!

> `nixpkgs` doesn’t define an option to configure socket dir for `postgres`, so we can consider renaming that for uniformity. Even `uds` in `process-compose-flake` can be re-considered.

I am not sure what the postgres config file instruction for this is called, but I guess using this would make sense. Maybe this even is `socketDir`. Process compose seems to be a bit messy as the CLI flags are `--use-uds` and `--unix-socket` but the environment variable for the latter is `PC_SOCKET_PATH`.

---

Overall, I don't really mind, all options are somewhat okay to find. Since different programs have different ways of offering the same functionality, some heterogeneity in the naming here is probably unavoidable. I suppose finding a unified way that intuitively makes sense is not as easy as I initially made it out to be. Good documentation can probably make any naming scheme work

PR #355, opened 2024-10-09
feat: Provide a `devShell` with packages from enabled services

### LINKED ISSUE #194: Have services provide the relevant devShell?
For eg., the postgres service could provide a devShell output that includes `psql` cli.

Ref (from an internal project that @shivaraj-bh should have access to):

image

Can we write that as:

```nix
pkgs.mkShell {
  inputsFrom = {
    config.devShells.common
    config.haskellProjects.default.outputs.devShell
    config.process-compose.services.services.devShell
  };
}
```

?

Because, *hiding the details* in `flake.nix` is generally a good thing.

shivaraj-bh:
resolves #194

### DIFF
--- a/nix/services/default.nix
+++ b/nix/services/default.nix
@@ -1,3 +1,4 @@
+{ pkgs, lib, config, ... }:
 let
   inherit (import ../lib.nix) multiService;
 in
@@ -24,4 +25,30 @@ in
     ./searxng.nix
     ./tika.nix
   ];
+
+  options.services.outputs.devShell = lib.mkOption {
+    type = lib.types.package;
+    readOnly = true;
+    description = ''
+      The devShell that aggregates packages from all the enabled services.
+    '';
+  };
+
+  config = {
+    services.outputs.devShell = pkgs.mkShell {
+      packages =
+        lib.pipe config.services [
+          # `outputs` is a reserved attribute set and is not the name of a service.
+          (lib.filterAttrs (n: _: n != "outputs"))
+          # Flatten services attrset
+          #
+          # Example:
+          # Input = { mysql."m1" = <cfg1>; mysql."m2" = <cfg2>; redis."r1" = <cfg3>; }
+          # Output = [ <cfg1> <cfg2> <cfg3> ]
+          (lib.foldlAttrs (acc: _: v: (lib.attrValues v) ++ acc) [ ])
+          (lib.filter (service: service.enable))
+          (map (service: service.package))
+        ];
+    };
+  };
 }
--- a/nix/services/default.nix
+++ b/nix/services/default.nix
@@ -1,9 +1,8 @@
-{ pkgs, lib, config, ... }:
 let
   inherit (import ../lib.nix) multiService;
 in
 {
-  imports = builtins.map multiService [
+  imports = (builtins.map multiService [
     ./apache-kafka.nix
     ./clickhouse
     ./elasticsearch.nix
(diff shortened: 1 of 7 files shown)


nix/services/default.nix
srid, line 1: As you can see, this file only imports other modules, without having any implementation of its own. So you can want to follow the convention by moving the options/config to `./nix/services/devshell.nix`.

nix/services/default.nix
```diff
@@ -24,4 +25,30 @@ in
...
+          # `outputs` is a reserved attribute set and is not the name of a service.
+          (lib.filterAttrs (n: _: n != "outputs"))
+          # Flatten services attrset
+          #
+          # Example:
+          # Input = { mysql."m1" = <cfg1>; mysql."m2" = <cfg2>; redis."r1" = <cfg3>; }
+          # Output = [ <cfg1> <cfg2> <cfg3> ]
+          (lib.foldlAttrs (acc: _: v: (lib.attrValues v) ++ acc) [ ])
+          (lib.filter (service: service.enable))
+          (map (service: service.package))
```
srid, line 50: This assumes that all services -- and services to be added in future -- have `package` option, right?

nix/services/default.nix
```diff
@@ -24,4 +25,30 @@ in
...
+        lib.pipe config.services [
+          # `outputs` is a reserved attribute set and is not the name of a service.
+          (lib.filterAttrs (n: _: n != "outputs"))
+          # Flatten services attrset
+          #
+          # Example:
+          # Input = { mysql."m1" = <cfg1>; mysql."m2" = <cfg2>; redis."r1" = <cfg3>; }
+          # Output = [ <cfg1> <cfg2> <cfg3> ]
+          (lib.foldlAttrs (acc: _: v: (lib.attrValues v) ++ acc) [ ])
+          (lib.filter (service: service.enable))
```
srid, line 49: You can possibly move this to `services.outputs.enabledServices` option, thus simplifying the code here.

srid (commented):
LGTM, with these changes.

nix/services/default.nix
srid: This could just be `listOf string`, fwiw.

nix/services/default.nix
shivaraj-bh: Yes

nix/services/default.nix
shivaraj-bh: https://github.com/juspay/services-flake/pull/355/commits/72abbb7d91c27ec3a7602dbda262570c46b8be8e

nix/services/default.nix
shivaraj-bh: An error is thrown if a service doesn’t define `package` option, since https://github.com/juspay/services-flake/pull/355/commits/72abbb7d91c27ec3a7602dbda262570c46b8be8e. 

We can safely assume every service will have a package, if there is an exception, we can revisit this in the future.

nix/services/devshell.nix
```diff
@@ -0,0 +1,53 @@
...
+in
+{
+  options.services.outputs = {
+    enabledServices = lib.mkOption {
+      type = types.listOf (types.listOf types.str);
+      readOnly = true;
+      description = ''
+        List of names of enabled services.
+
+        Each item will be of the form [ "<service>" "<instance-name>" ]
```
srid, line 13: Why not `{ service = "<service">; instance = "<instance-name>"; }`? Or even `{ service1 = [ instance 1 ... ]; ... }`.

Types should not be designed based on the logic of how they get used (`attrByPath`), but rather by clarity of purpose. Also the lists can represent illegal shapes (`["foo" "bar" "qux"]`)  than a structured attrset here.

srid (commented):
LGTM, can merge after this type change.

nix/services/devshell.nix
shivaraj-bh: How about https://github.com/juspay/services-flake/pull/355/commits/b92546c0c823180c03e3a331594e5cdf9d05d8fe?

srid (approved):

PR #356, opened 2024-10-09
feat: Add options.declarativePlugins to grafana

conscious-puppet:


nix/services/grafana.nix
```diff
@@ -71,6 +71,17 @@ in
       '';
     };
 
+    declarativePlugins = lib.mkOption {
```
srid, line 74: Perhaps just `plugins`? As there's no way to set things imperatively in flakes anyway.

conscious-puppet: i tried to have the same option as in the nixos services modules. 

ref: https://github.com/NixOS/nixpkgs/blob/dd8f501ee76c12ae47aa5e010827230afaf21eaf/nixos/modules/services/monitoring/grafana.nix#L234

srid: Makes sense then.

PR #372, opened 2024-10-21
chore(example/without-flake-parts): init

shivaraj-bh:
A useful addition that demonstrates more ways to use `services-flake`


srid:
Cool, this might be of interest to @vancoding too.

PR #385, opened 2024-11-05
fix(clickhouse-init): kill `clickhouse-server` on `EXIT`

### LINKED ISSUE #361: process-compose gets stuck on quit if clickhouse init db fails
process-compose gets stuck on quit if the the clickhouse init db service fails. we need to manually kill the clickhouse process.

the issue can be reproduced by using a malformed initial db dump with the clickhouse service.

[asciicast](https://asciinema.org/a/dEXn6hpILm7NFaL5kalVn6yRi)

shivaraj-bh:
fixes #361

### DIFF
--- a/nix/services/clickhouse/bad_test.sql
+++ b/nix/services/clickhouse/bad_test.sql
@@ -1,0 +1,2 @@
+CREATE STABLE sample_db.ride (`id` Int64, `short_id` String) ENGINE = MergeTree() PRIMARY KEY (id);
+
--- a/nix/services/clickhouse/clickhouse_test.nix
+++ b/nix/services/clickhouse/clickhouse_test.nix
@@ -1,4 +1,4 @@
-{ pkgs, config, ... }: {
+{ pkgs, config, name, ... }: {
   services.clickhouse."clickhouse1" = {
     enable = true;
     port = 9000;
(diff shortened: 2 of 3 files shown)


shivaraj-bh:
@conscious-puppet can you check if this PR solves your issue?

conscious-puppet:
just a suggestion, may be we can use this process compose option to prevent process-compose shutdown automatically

```
      --keep-project            keep the project running even after all processes exit
```

ref: https://f1bonacc1.github.io/process-compose/cli/process-compose/

shivaraj-bh:
> just a suggestion, may be we can use this process compose option to prevent process-compose shutdown automatically
> 
> ```
>       --keep-project            keep the project running even after all processes exit
> ```
> 
> ref: https://f1bonacc1.github.io/process-compose/cli/process-compose/

You can always do `nix run .#<pc-project-name> -- --keep-project`. Otherwise, there is also an option to run with `--tui=false` mode, so that the logs are dumped on the terminal’s stdout.

To configure in nix to use `--keep-project` always will require contribution upstream in https://github.com/Platonic-Systems/process-compose-flake/blob/018783e68c89f94dbe11ce918d6206e9e4138787/nix/process-compose/cli.nix#L31-L73. I would also propose to add one option `extraCliArgs` for such args instead of creating a specific option for it.

PR #393, opened 2024-11-19
chore(nix): Remove treefmt

srid:
We already use pre-commit and that supports nixpkgs-fmt.

cf. https://github.com/srid/haskell-template/pull/144

Also, this PR updates git-hooks.nix while making a non-flake thereby reducing the number of lock file (`dev/flake.lock`) entries.

### DIFF
--- a/dev/flake.lock
+++ b/dev/flake.lock
@@ -15,22 +15,6 @@
         "type": "github"
       }
     },
-    "flake-compat": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1696426674,
-        "narHash": "sha256-kvjfFW7WAETZlt09AgDn1MrtKzP7t90Vf7vypd3OL1U=",
-        "owner": "edolstra",
-        "repo": "flake-compat",
-        "rev": "0f9255e01c2351cc7d116c072cb317785dd33b33",
-        "type": "github"
-      },
-      "original": {
-        "owner": "edolstra",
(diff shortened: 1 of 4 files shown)


shivaraj-bh:
The stdout when there is a modification to be done using `pre-commit`:
```sh
services-flake on  main [$!] via ❄️  impure (nix-shell-env) took 6s
❯ just fmt
pre-commit run -a
nixpkgs-fmt..............................................................Failed
- hook id: nixpkgs-fmt
- files were modified by this hook

flake.nix
1 / 8 have been reformatted
0 / 8 have been reformatted
0 / 8 have been reformatted
0 / 8 have been reformatted
0 / 8 have been reformatted
0 / 8 have been reformatted
0 / 8 have been reformatted
0 / 3 have been reformatted

error: Recipe `fmt` failed on line 29 with exit code 1
```

The `error` at the end is a bit misleading because it makes one think that formatting didn’t run successfully.

srid:
I should have thought more on this.

pre-commit does not have a notion of "run all hooks unconditionally". Let's just get rid of the `fmt` target entirely, and briefly document pre-commit in the README like haskell-template now does:

https://github.com/srid/haskell-template/commit/b224c24f7db6afea775d80a994c12ded162ecbf4

Besides, `fmt` naming is now inaccurate anyway - since you can do more than formatting (e.g.: run hlint).

PR #443, opened 2025-02-05
chore(test): disable elasticsearch on darwin

shivaraj-bh:
Fails on macOS with: 
```sh
error: chmod '"/nix/store/rcx3n94ygmd61rrv2p22sykhk0yx49n4-elasticsearch-7.17.16/modules/x-pack-ml/platform/darwin-aarch64/controller.app"': Operation not permitted
```
Related: https://github.com/NixOS/nix/issues/6765

### DIFF
--- a/test/flake.nix
+++ b/test/flake.nix
@@ -42,7 +42,6 @@
             "${inputs.services-flake}/nix/services/apache-kafka_test.nix"
             "${inputs.services-flake}/nix/services/cassandra_test.nix"
             "${inputs.services-flake}/nix/services/clickhouse/clickhouse_test.nix"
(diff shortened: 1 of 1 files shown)


tomberek:
Do you build this on a machine with/without Xcode?

shivaraj-bh:
> Do you build this on a machine with/without Xcode?

Oh, I missed the notification for this comment.

The macOS runner seems to have the Xcode command line tools but not the whole deal because when I run `xcodebuild` I get:
```sh
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
```

PR #448, opened 2025-02-14
feat: Portable services

shivaraj-bh:
resolves #354 

Init with `ollama`.

### process-compose app

```nix
# Inside perSystem
{
  process-compose.sf-ollama = {
    imports = [ inputs.services-flake.processComposeModules.default ];
    services.ollama.”o1”.enable = true;
  };
}
```
And `nix run .#sf-ollama` to start the service

### `systemd`/`launchd` user unit/agent

> Note: The option `services.<service-name>` overlaps with existing service definitions in `home-manager`. I could rename it to avoid conflict, but for now you could use the `release-24.11` branch of `home-manager` that doesn’t yet have `ollama`. 

```nix
# In your `homeConfiguration`
{
  imports = [ inputs.services-flake.homeModules.default ];
  services.ollama.”o1”.enable = true;
}
```

Lastly, activate your home-manager config...

<Insert home-manager template to test `Ollama` running via `systemd` or `launchd`>

TODO:
- [ ] `outputs.{systemd, launchd, process-compose}` must share as much configuration as possible (also add sane defaults for `systemd` and `launchd`)
- [ ] `systemd` and `launchd` tests
- [ ] Documentation

PR #466, opened 2025-02-23
feat: Add process-compose `defaults` module

### LINKED ISSUE #459: Wish: Disable the process-compose HTTP server by default
Hey,

related to #222 and general usage.
Process compose will now by default bind to :8080 which is in local dev environments is a very popular port.

So I would recommend that `cli.options.no-server = true;` will be set by default, instead of repeating this for every individual service.

From #222
```
nix run "github:juspay/services-flake?dir=example/simple"
```
I assume this would still work today, if process compose wouldn't bind to :8080. Unless there is a future interaction planned between services-flake and process-compose.

Thanks for your consideration.

shivaraj-bh:
resolves #459

### DIFF
--- a/example/llm/flake.nix
+++ b/example/llm/flake.nix
@@ -19,10 +19,6 @@
         imports = [
           inputs.services-flake.processComposeModules.default
         ];
-        cli = {
-          # Disable the process-compose HTTP server
-          options.no-server = true;
-        };
         services = let dataDirBase = "$HOME/.services-flake/llm"; in {
           # Backend service to perform inference on LLM models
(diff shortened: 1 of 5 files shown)


shivaraj-bh:
This PR also lays ground work by structuring some `process-compose` specific definitions under `/nix/process-compose`, paving way for better structuring multiple backend definitions in https://github.com/juspay/services-flake/pull/448

srid (approved):
LGTM.

If you are looking to re-arrange files under `./nix/process-compose` I suggest doing that in a PR and merging it (single commit) before further work.

Issue #485, opened 2025-03-19
`mysql` configure script fails to run


greg-hellings:
The default-configure step fails to run with `ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)` as its output message. The configuration for the script is

```nix
mysql.default = {
  enable = true;
  dataDir = "./.data/mysql";
  initialDatabases = [ { name = "django"; } ];
  ensureUsers = [ {
    password = "django_pw";
    name = "django";
    ensurePermissions."*.*" = "ALL PRIVILEGES";
  } ];
};
```

It makes sense that the script cannot connect over the socket in `/run/mysqld/mysqld.sock` because there is no such server running. However, there is an existing file at `/etc/mysql/my.cnf` that includes the snippet

```my.cnf
[client-server]
socket = /run/mysqld/mysqld.sock
```

It seems that, by default, the mysql client commands in the configure script are picking up that file and incorrectly trying to find a globally running server rather than the process-compose instance.

shivaraj-bh:
https://github.com/juspay/services-flake/pull/486 does this fix it?

greg-hellings:
It does look similar to (and better than) the quick workaround I had put in place to test where exactly the error was coming from. I just added the `--socket` argument directly to those same lines. In the end I just moved the `my.cnf` file out of the way as it was the generic one from the OS package manager (system76 host) so I didn't need to use a patched flake.

If you want a firm answer I can give you one next week when I'm back on the work laptop.

PR #486, opened 2025-03-19
fix(mysql): Explicitly pass `my.cnf` path to avoid global fallback

shivaraj-bh:
resolves https://github.com/juspay/services-flake/issues/485


shivaraj-bh:
I have tested the scenario with a global `my.cnf` file and it works despite its presence

PR #514, opened 2025-05-05
fix(mysql): allow hyphens in database names

### LINKED ISSUE #513: mysql service does not allow database names with "-" in it
```nix
services.mysql."mydatabase" = {
  enable = true;
  initialDatabases = [{ name = "test-database"; }];
}
```
does not work.

puzzleshark:
MySQL service rejected DB names containing “-”, although MariaDB allows them. Fixes #513


shivaraj-bh:
Thanks for the contribution!

Issue #526, opened 2025-05-28
bug: mongodb fails to start when using non-default port


adrian-gierakowski:
The reason is that the port is not being used by the [probe command](https://github.com/juspay/services-flake/blob/8bc6dff1c0d82842b28e3906ac4645a3c3a49dbe/nix/services/mongodb.nix#L84) (as well as the init command, however due to failing probe it never gets executed anyway)

I've got [an implementation](https://github.com/adrian-gierakowski/services-flake/commit/fe1df6753f6ded7dffb5b6a9bf95be3455d2474c) of a mongodb services I've been using privately, from which I would like to port missing features (like ability to run a single-member replica set, which is needed to use [change streams](https://www.mongodb.com/docs/manual/changeStreams/)) to this repo. I can fix this bug while I'm at it.

PR #528, opened 2025-05-28
feat(mongodb): allow running as single member replicat set

adrian-gierakowski:
this is needed to use change streams

also
- add ulimit config, which can be used to set nubmer of open descritors
  which mongod is allow to use
- fix probe not respecting host\port config (https://github.com/juspay/services-flake/issues/526)

### DIFF
--- a/nix/services/mongodb.nix
+++ b/nix/services/mongodb.nix
@@ -48,86 +48,144 @@ in
       default = "";
       description = "Additional text to be appended to `mongodb.conf`.";
     };
-  };
 
-  config = {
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
Haven't thought about different files for tests yet. That would be quite useful when the number of instances of service keeps growing. Moreover, since all these instances define their test in the same "test" process, it can get cluttered quite soon.


How about we use ".nix/services/mongodb_<test-name>_test.nix"? We could standardise/improve it later when more such test files start appearing.

PR #533, opened 2025-06-06
fix(postgres): Quote `unix_socket_directories` argument

### LINKED ISSUE #532: postgres: `init` process fails when there is a space in the absolute path of the data directory
Observed this behaviour on macOS.

Might need to handle space in absolute path in https://github.com/juspay/services-flake/blob/491f3b5a35d8da0149f4bebfed7c2c36a185fa5b/nix/services/postgres/setup-script.nix#L116

niksingh710:
Wrapped `PGHOST` with `"` for `unix_socket_directories.

Resolves #532.
| Initial | After |
|---------|-------|
| Screenshot 2025-06-06 at 15 09 18 | Screenshot 2025-06-06 at 15 12 47 |

In the Initial image the error is because of the asbsolute path getting broken into 2 argsuments.
```sh
pg_ctl -D ./data/pg2 -w start -o '-c unix_socket_directories=/Users/nikhil.singh/work/services flake/test/test/new/ socket/path/pg-init-m2GRBV -c listen_addresses= -p 5433'
waiting for server to start....postgres: invalid argument: "flake/test/test/new/socket/path/pg-init-m2GRBV"
Try "postgres --help" for more information.
```

Where as in After it states that 
```sh
Listening on Unix socket "/Users/nikhil.singh/work/services flake/test/test/new/socket/path3/.s.PGSQL.5432"
```

### DIFF
--- a/nix/services/postgres/setup-script.nix
+++ b/nix/services/postgres/setup-script.nix
@@ -129,7 +129,7 @@ in
 
       trap 'pg_ctl -D "$PGDATA" -m fast -w stop' EXIT
 
-      pg_ctl -D "$PGDATA" -w start -o "-c unix_socket_directories=$PGHOST -c listen_addresses= -p ${toString config.port}"
+      pg_ctl -D "$PGDATA" -w start -o "-c unix_socket_directories=\"$PGHOST\" -c listen_addresses= -p ${toString config.port}"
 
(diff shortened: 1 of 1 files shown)


shivaraj-bh (approved):
Thanks

PR #545, opened 2025-07-01
feat: Add global settings option to phpfpm

john-shaffer:
Add a "settings" option to the phpfpm module.
This allows changing global settings such as
the logging level.

The existing "extraConfig" option sets the pool
settings, but cannot set global settings.

The option name "settings" is taken from the
NixOS Module.

### DIFF
--- a/nix/services/phpfpm.nix
+++ b/nix/services/phpfpm.nix
@@ -83,6 +83,24 @@ in
         }
       '';
     };
+
+    globalSettings = lib.mkOption {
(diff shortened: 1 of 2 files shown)


terlar:
With this probably the daemonize and error_log should be converted into defaults that is merged with this new thing. That is also what the nixos module does, but it seemed a bit unnecessarily complicated, but perhaps that would be needed after all

shivaraj-bh:
Thanks

PR #547, opened 2025-07-03
fix(phpfpm): Health checks fails when env is large

john-shaffer:
When the total environment is too large, such as
over 8KB, FastCGI closes the connection with
no output or log messages. cgi-fcgi returns with
an exit code of 104 and no output.

This happens even when clear_env is set in the
PHP-FPM config, since the problem seems to happen
before that would be applied.

To fix this, we prefix the cgi-fcgi command with
`env -i` to clear the environment.

### DIFF
--- a/nix/services/phpfpm.nix
+++ b/nix/services/phpfpm.nix
@@ -154,11 +154,14 @@ in
                   config.listen;
             in
             {
+              # php-fpm can ignore requests that are too large, so we
+              # use env -i to remove all environment variables.
+              # See https://github.com/php/php-src/issues/16042
               exec.command =
                 if (builtins.isInt config.listen) then
-                  "${pkgs.fcgi}/bin/cgi-fcgi -bind -connect 127.0.0.1:${toString config.listen}"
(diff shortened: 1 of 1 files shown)


terlar:
Thank you for fixing this, I did run into this myself but didn't figure out what the root cause was. Glad to have this fix for now I just disabled the check when I ran into this.

shivaraj-bh:
How can I reproduce this?

john-shaffer:
`nix develop ./dev -c just run phpfpm` reproduces the failure from within the services-flake repo because nix creates an environment that is over 15KiB. If for some reason that doesn't work, you can force it with  `BIGVAR=$(head -c 8192 < /dev/zero | tr '\0' 'A') nix develop ./dev -c just run phpfpm`.

shivaraj-bh:
Thanks.

Yes, I can reproduce with `just run phpfpm` on the base branch.

The workaround looks good to me. Can you add a comment above the check for posterity? Also, maybe a link to the source for the environment limit, if any.

john-shaffer:
Added comments with link to https://github.com/php/php-src/issues/16042 which seems to be the root cause of the problem.

PR #574, opened 2025-08-15
feat(pubsub-emulator): init service

KDanisme:
its a bit weird because there is no pubsub emulator package, but a component from google-clioud-sdk pacakge

thats the only part I wasn't sure how to cleanup

### DIFF
--- a/doc/pubsub-emulator.md
+++ b/doc/pubsub-emulator.md
@@ -1,0 +1,8 @@
+
+# Pubsub Emulator
+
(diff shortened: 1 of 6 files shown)


KDanisme:
Review resolved

shivaraj-bh:
Re https://github.com/juspay/services-flake/pull/574/commits/a3d88426fcf878d976a07291afd7c2424433b688: I need to document it in https://community.flake.parts/services-flake/contributing#new-service. There are also some services that don’t follow that order (esp. in https://github.com/juspay/services-flake/blob/71c69533bdae75f7414d348c9d6878436cffcd98/nix/services/default.nix#L6-L30), I will fix them and document in a different PR.

PR #610, opened 2025-10-22
fix(postgres): quote `dataDir` to prevent word splitting

semi710:
Escape seq fix, similar to #533


shivaraj-bh:
Looks good.

I wonder if there is any automatic way to detect these things. shellcheck works by recommending to quote env vars, but it wouldn’t work for interpolated strings (like `dataDir` in this case)

PR #617, opened 2025-10-31
feat(multiService): Add service option to modify underlying process settings

### LINKED ISSUE #616: multiService: Add `depends_on` as a default option
`multiService` defines a few [options defaults](https://github.com/juspay/services-flake/blob/4104b44868206c7fd7738e349d814a352c480711/nix/lib.nix#L18-L30) like `enable`, `namespace` and `dataDir`, which can reasonably expected to be defined by all the services.

I propose to also include `depends_on` here.

## Why?

- Often we end up splitting `depends_on` config for a service away from the service config, see

  ```nix
  {
    open-webui."open-webui1" = {
       enable = true;
        …
    };

    # Start the Open WebUI service after the Ollama service has finished initializing and loading the models
     settings.processes.open-webui1.depends_on.ollama1-models.condition = "process_completed_successfully";
  }
  ```
  [(code snippet from example/llm)](https://github.com/juspay/services-flake/blob/4104b44868206c7fd7738e349d814a352c480711/example/llm/flake.nix#L49-L75)

  If this proposal is implemented, it can be changed to:
  ```nix
  {
    open-webui."open-webui1" = {
       enable = true;
       depends_on.ollama1-models.condition = "process_completed_successfully";
        …
    };
  }
  ``` 

- `depends_on` is an option almost always used while defining complex service dependency relationship.

shivaraj-bh:
resolves #616

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -25,6 +25,9 @@
             type = lib.types.lazyAttrsOf lib.types.deferredModule;
             description = "Settings for a process-compose process defined in this service";
             default = { };
+            apply = v: {
+              processes = v;
+            };
           };
           outputs = {
             defaultProcessSettings = lib.mkOption {
@@ -45,12 +48,11 @@
                 process-compose settings for the processes under the ${service} service
               '';
               apply = v: v // {
-                processes = lib.flip lib.mapAttrs v.processes (pName: cfg:
+                processes = lib.flip lib.mapAttrs v.processes (_: cfg:
                   {
                     imports = [
                       config.outputs.defaultProcessSettings
                       cfg
-                      config.processSettings.${pName}
                     ];
                   }
                 );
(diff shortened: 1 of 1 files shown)


nix/lib.nix
```diff
@@ -21,12 +21,10 @@
...
             description = "The directory where all data for `${service}.<name>` is stored";
           };
-          namespace = lib.mkOption {
-            description = ''
-              Namespace for the ${service} service
-            '';
-            default = "${service}.${name}";
-            type = lib.types.str;
+          processSettings = lib.mkOption {
+            type = lib.types.lazyAttrsOf lib.types.deferredModule;
```
shivaraj-bh: lazyAttrsOf because a service can define multiple processes (postgres, postgres-init for example)

PR #639, opened 2026-02-05
dynamodb-local: init

shunueda:
Adding dynamodb-local, an official, local version of Dynamo DB. 

Healthcheck is awkward, but following the pattern from [devenv](https://github.com/cachix/devenv/blob/b91e1b24ff6f7bddcdbb00e5834ed457a4126bf2/src/modules/services/dynamodb-local.nix#L63) and [localstack](https://github.com/localstack/localstack/blob/cf9a6fcf0e2c3ee7daebe5a8d540148a63f9490a/localstack-core/localstack/services/dynamodb/server.py#L216).

### DIFF
--- a/nix/services/dynamodb-local.nix
+++ b/nix/services/dynamodb-local.nix
@@ -1,4 +1,10 @@
-{ pkgs, lib, name, config, ... }: {
+{ pkgs
+, lib
+, name
+, config
+, ...
+}:
+{
   options = {
     package = lib.mkPackageOption pkgs "dynamodb-local" { };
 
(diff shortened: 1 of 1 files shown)


nix/services/dynamodb-local.nix
```diff
@@ -0,0 +1,97 @@
...
+      startCommand = pkgs.writeShellApplication {
+        name = "start-dynamodb";
+        runtimeInputs = [ config.package ];
+        text = ''
+          ${lib.optionalString (config.dbPath != null) ''
+            mkdir -p ${config.dbPath}
+          ''}
+           exec dynamodb-local \
+             -port ${toString config.port} \
+             ${lib.optionalString (config.dbPath != null) "-dbPath ${config.dbPath}"} \
```
shivaraj-bh:
```suggestion
             ${lib.optionalString (config.dbPath != null) "-dbPath ${lib.escapeShellArg config.dbPath}"} \
```

to prevent any word splitting or globbing issues

PR #647, opened 2026-03-10
fix(dynamodb-local): fix check on inMemory to allow setting dbPath

attilaersek:
What changed?
Adjusted the check in the dynamodb-local service so that the inMemory flag and dbPath setting are evaluated correctly.
Added appropriate test(s) to cover both modes and ensure dbPath can be set when inMemory = false.

### DIFF
--- a/nix/services/dynamodb-local.nix
+++ b/nix/services/dynamodb-local.nix
@@ -34,8 +34,8 @@
       '';
       apply =
         v:
-        lib.throwIf (config.dbPath != null) ''
-          You can't specify both -dbPath and -inMemory at once.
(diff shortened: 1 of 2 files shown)


shivaraj-bh:
Not a concern of this PR, but it occurred to me while reviewing that: asserting in `inMemory` option with (`dbPath`) cross-option reference felt a bit awkward. Maybe a NixOS module style `assertions` option ([example](https://github.com/NixOS/nixpkgs/blob/5adf8456b4533241382863c17753b8c8b8b83210/nixos/modules/config/ldso.nix#L47-L53)) would be nice in such cases. Though straightforward, not enough cases to add support yet; I will re-visit later.

PR #656, opened 2026-03-28
Add Keycloak

johnhamelink:
Implement Keycloak service based on the [devenv implementation](https://github.com/cachix/devenv/blob/main/src/modules/services/keycloak.nix#LL456). Feedback very much appreciated - this is my first time contributing to services-flake!

## TODOs:
- [x] Running Keycloak Service
- [x] Working Readiness Check
- [x] Passing test
- [x] Realm import
- [ ] Realm export
- [x] Rigorous user testing
- [x] Fix all outstanding FIXMEs/TODOs in comments

image

---

## Questions:

1. The [devenv implementation comes with a series of scripts](https://github.com/cachix/devenv/blob/main/src/modules/services/keycloak.nix#L538): how should I implement them in services-flake?
2. The devenv implementation also [runs a script after termination of the service](https://github.com/cachix/devenv/blob/main/src/modules/services/keycloak.nix#L554). I'm also unsure how to implement this?

### DIFF
--- a/nix/services/default.nix
+++ b/nix/services/default.nix
@@ -33,6 +33,7 @@ in
     ./loki.nix
     ./phpfpm.nix
     ./pubsub-emulator.nix
+    ./keycloak.nix
   ]) ++ [
     ./devshell.nix
   ];
--- a/nix/services/keycloak.nix
+++ b/nix/services/keycloak.nix
@@ -108,6 +108,7 @@ in
                   inStore = false;
                   absolute = false;
                 }
+                # types.path
               );
               default = null;
               example = "./realms/a.json";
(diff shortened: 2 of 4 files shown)


johnhamelink:
I've been using this to work on a project for a while now and it seems to work well for me!

shivaraj-bh:
> The [devenv implementation comes with a series of scripts](https://github.com/cachix/devenv/blob/main/src/modules/services/keycloak.nix#L538): how should I implement them in services-flake?

We don’t export scripts in services-flake. it’d be nice to add documentation for how users can define their own scripts by reading the values for the keycloak service’s options. Happy to help with this if it isn’t clear.

> The devenv implementation also [runs a script after termination of the service](https://github.com/cachix/devenv/blob/main/src/modules/services/keycloak.nix#L554). I'm also unsure how to implement this?

process-compose supports `depends_on` for defining any process that should run conditional to the state of some other process. Example: https://github.com/juspay/services-flake/blob/684700114e37d4335459174cf7213552c85c1a58/nix/services/minio.nix#L122. I haven’t checked what exactly the script in the devenv service does, i will do that once I sit to review the PR thoroughly.

shivaraj-bh:
I got occupied with other things. I will spend some time on the PR this week

gabyx:
@shivaraj-bh:
Thanks for pushing this.

As I was the maintainer of the keycloak devenv module and I did not see this PR already.
I opened another which is basically finished and has all tests since we also needed that one.
See #700. I would close this one probably if there is not a new feature which was added in your PR.

My PR replicates the keycloak service as in devenv.

PR #688, opened 2026-05-25
grafana-pyroscope: init

kashw2:
Adds Grafana Pyroscope as a service

https://grafana.com/oss/pyroscope/

### DIFF
--- a/nix/services/pyroscope.nix
+++ b/nix/services/pyroscope.nix
@@ -1,0 +1,105 @@
+{ pkgs
+, lib
+, name
+, config
+, ...
+}:
+let
+  inherit (lib) types;
+  yamlFormat = pkgs.formats.yaml { };
+in
+{
+  options = {
+    package = lib.mkPackageOption pkgs "pyroscope" { };
+
+    httpAddress = lib.mkOption {
+      type = types.str;
+      description = "Which address to access pyroscope from.";
+      default = "127.0.0.1";
+    };
+
+    httpPort = lib.mkOption {
+      type = types.port;
+      description = "Which port to run pyroscope on.";
+      default = 4040;
+    };
+
+    extraConfig = lib.mkOption {
+      type = yamlFormat.type;
+      default = { };
+      description = ''
+        Specify the configuration for Pyroscope in Nix.
+
+        See https://grafana.com/docs/pyroscope/latest/configure-server/reference-configuration-parameters/ for available options.
+      '';
+    };
+    extraFlags = lib.mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description = "Additional flags to pass to pyroscope.";
+    };
+  };
+
+  config = {
+    outputs = {
+      settings = {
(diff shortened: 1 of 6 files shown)


nix/services/pyroscope.nix
```diff
@@ -0,0 +1,97 @@
...
+                  };
+                };
+              }
+              config.extraConfig;
+            pyroscopeConfigYaml = yamlFormat.generate "pyroscope.yaml" pyroscopeConfig;
+            startScript = pkgs.writeShellApplication {
+              name = "start-pyroscope";
+              runtimeInputs = [ config.package ];
+              text = ''
+                pyroscope --config.file=${pyroscopeConfigYaml} ${lib.escapeShellArgs config.extraFlags}
```
shivaraj-bh:
```suggestion
                exec pyroscope --config.file=${pyroscopeConfigYaml} ${lib.escapeShellArgs config.extraFlags}
```

nix/services/pyroscope_test.nix
```diff
@@ -0,0 +1,36 @@
...
+    extraFlags = [
+      "-ingester.lifecycler.addr=127.0.0.1"
+      "-distributor.ring.instance-addr=127.0.0.1"
+      "-compactor.ring.instance-addr=127.0.0.1"
+      "-overrides-exporter.ring.instance-addr=127.0.0.1"
+      "-query-scheduler.ring.instance-addr=127.0.0.1"
+      "-query-frontend.instance-addr=127.0.0.1"
+      "-store-gateway.sharding-ring.instance-addr=127.0.0.1"
+      "-memberlist.bind-addr=127.0.0.1"
+    ];
```
shivaraj-bh: I wonder if it is a good idea to make these default (not hardcode 127.0.0.1, but use httpAddress) in the `pyroscopeConfig` that is merged with `extraConfig` (in nix/services/pyroscope.nix)?

kashw2: This sounds like a reasonable default

nix/services/pyroscope.nix
kashw2: Done in 1c60fc45ce60c4fac5bfdc8fdad6fa21adc2580b. Out of curiosity, is there a reason we couldn't `lib.getExe cfg.package ...` here as well? My assumption is that it's because it'd be a bit semi redundant to have a `runtimeInputs` containing the package just to use `lib.getExe` in the `text` of the shell application.

nix/services/pyroscope_test.nix
kashw2: Done in 1c60fc45ce60c4fac5bfdc8fdad6fa21adc2580b

nix/services/pyroscope.nix
shivaraj-bh: no reason. feel free to use it

shivaraj-bh:
ignore the aarch64-darwin CI failure, not related to this PR

PR #689, opened 2026-05-30
seaweedfs: init

kashw2:
Adds SeaweedFS. A MinIO alternative

### DIFF
--- a/nix/services/seaweedfs.nix
+++ b/nix/services/seaweedfs.nix
@@ -1,0 +1,187 @@
+{ pkgs
+, lib
+, name
+, config
+, ...
+}:
+let
+  inherit (lib) types;
+in
+{
+  options = {
+    package = lib.mkPackageOption pkgs "seaweedfs" { };
+    host = lib.mkOption {
+      type = types.str;
+      default = "127.0.0.1";
+      description = ''
+        IP or hostname the SeaweedFS server binds to and advertises as.
+      '';
+    };
+    master = {
+      port = lib.mkOption {
+        type = types.port;
+        default = 9333;
+        description = "HTTP port for the master server.";
+      };
+      grpcPort = lib.mkOption {
+        type = types.nullOr types.port;
+        default = null;
+        description = ''
+          gRPC port for the master server. If null, defaults to `master.port + 10000`.
+        '';
+      };
+      defaultReplication = lib.mkOption {
+        type = types.str;
+        default = "000";
+        description = ''
+          Default replication policy. `000` disables replication
+        '';
+      };
+      volumeSizeLimitMB = lib.mkOption {
+        type = types.ints.positive;
+        default = 1024;
+        description = "Master stops directing writes to volumes larger than this.";
+      };
+    };
+    volume = {
+      port = lib.mkOption {
+        type = types.port;
+        default = 8080;
+        description = "HTTP port for the volume server.";
+      };
+      grpcPort = lib.mkOption {
+        type = types.nullOr types.port;
+        default = null;
+        description = ''
+          gRPC port for the volume server. If null, defaults to `volume.port + 10000`.
+        '';
+      };
+      max = lib.mkOption {
+        type = types.str;
+        default = "8";
+        description = ''
+          Maximum number of volumes for the volume server.
+        '';
+      };
+    };
+    filer = {
+      enable = lib.mkEnableOption "the SeaweedFS filer server";
+      port = lib.mkOption {
+        type = types.port;
+        default = 8888;
+        description = "HTTP port for the filer server.";
+      };
+      grpcPort = lib.mkOption {
+        type = types.nullOr types.port;
+        default = null;
+        description = ''
+          gRPC port for the filer server. If null, defaults to `filer.port + 10000`.
+        '';
+      };
+    };
+    s3 = {
+      enable = lib.mkOption {
+        type = types.bool;
+        default = false;
+        description = "Whether to enable the S3 gateway.";
+        apply =
+          v:
+          lib.throwIf
+            (
+              v && !config.filer.enable
+            ) "services.seaweedfs.${name}.filer.enable must be true when s3.enable is true"
+            v;
+      };
+      port = lib.mkOption {
+        type = types.port;
+        default = 8333;
+        description = "HTTP port for the S3 API.";
+      };
+      config = lib.mkOption {
+        type = types.nullOr types.path;
+        default = null;
+        description = ''
+          Path to an S3 config file (identities/credentials). When null, the S3
+          gateway runs without authentication.
+        '';
+      };
+    };
+    extraArgs = lib.mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description = "Extra arguments to pass to `weed server`.";
+    };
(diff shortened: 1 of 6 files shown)


nix/services/seaweedfs_test.nix
```diff
@@ -0,0 +1,31 @@
...
+    in
+    {
+      command = pkgs.writeShellApplication {
+        name = "seaweedfs-test";
+        runtimeInputs = [ pkgs.curl ];
+        text = ''
+          # The master is gated by the readiness probe; the volume, filer, and s3 gateway each start a moment later, so retry on conn refused.
+          curl_with_retry() {
+            curl -fsS -o /dev/null --retry 2 --retry-delay 2 --retry-connrefused "$1"
+          }
```
shivaraj-bh: It’d be nicer to instead replace the `readiness_probe.http_get` with a `readiness_probe.exec.command` that runs curl’s to gate the readiness probe on all of them

nix/services/seaweedfs.nix
```diff
@@ -0,0 +1,168 @@
...
+      grpcPort = lib.mkOption {
+        type = types.nullOr types.port;
+        default = null;
+        description = ''
+          gRPC port for the filer server. If null, defaults to `filer.port + 10000`.
+        '';
+      };
+    };
+    s3 = {
+      enable = lib.mkEnableOption "the S3 gateway (requires filer)";
```
shivaraj-bh: instead of mentioning in description, I would assert it:

```nix
apply = v:
        lib.throwIf (v && !config.filer.enable)
          "services.seaweedfs.${name}.filer.enable must be true when s3.enable is true"
          v;
```

shivaraj-bh:
just a meta note: good to see an alternative since I had to https://github.com/juspay/services-flake/commit/64f5794e831ad8f80ff1f98424f3a1c047ba1df4

nix/services/seaweedfs_test.nix
kashw2: TIL I could exec multiple curl requests to test for readiness. I'd been looking for just this when working on this, thanks a bunch! Resolved in 91fe40a84c5800bde27a7ceef756392dd9e00a1e

nix/services/seaweedfs.nix
kashw2: Applied in 91fe40a84c5800bde27a7ceef756392dd9e00a1e

nix/services/seaweedfs.nix
```diff
@@ -0,0 +1,187 @@
...
+
+    readiness_probe = {
+      exec.command =
+        let
+          check = url: "${lib.getExe pkgs.curl} -fsS -o /dev/null ${lib.escapeShellArg url}";
+        in
+        # Below generates curl checks for all endpoints separated by &&
+        lib.concatStringsSep " && " (
+          [
+            (check "http://${config.host}:${toString config.master.port}/cluster/status")
```
shivaraj-bh: upstream e2e docker image uses `/cluster/healthz` and `/healthz` (for volume). https://github.com/seaweedfs/seaweedfs/blob/38a47d1dd39a024d6d32fe5a934c73d93b943859/docker/compose/e2e-mount.yml#L8

any reason `/status` is used here?

kashw2: In my original implementation I was trying to have more sophisticated tests that involved more than just testing if a request succeeded and checking responses with grep. It might have been something from then that continued to work after I abandoned that decision in favor of just testing against whether or not a request worked or not. I'll swap it to be inline with upstream

nix/services/seaweedfs.nix
```diff
@@ -0,0 +1,187 @@
...
+          check = url: "${lib.getExe pkgs.curl} -fsS -o /dev/null ${lib.escapeShellArg url}";
+        in
+        # Below generates curl checks for all endpoints separated by &&
+        lib.concatStringsSep " && " (
+          [
+            (check "http://${config.host}:${toString config.master.port}/cluster/healthz")
+            (check "http://${config.host}:${toString config.volume.port}/healthz")
+          ]
+          ++ lib.optional config.filer.enable (check "http://${config.host}:${toString config.filer.port}/")
+          ++ lib.optional config.s3.enable (check "http://${config.host}:${toString config.s3.port}/")
```
shivaraj-bh: is there no healthz endpoint like above?

I was wondering since I’d assume healthz is meant specifically for healtchecks

test/flake.nix
```diff
@@ -62,8 +62,8 @@
             "${inputs.services-flake}/nix/services/neo4j_test.nix"
             "${inputs.services-flake}/nix/services/redis_test.nix"
             "${inputs.services-flake}/nix/services/redis-cluster_test.nix"
+            "${inputs.services-flake}/nix/services/seaweedfs_test.nix"
```
shivaraj-bh: can you move this below where the test only runs on `x86_64-linux`?

it’d be nice if you can also add a comment saying “darwin build fixed in https://github.com/NixOS/nixpkgs/pull/534897"

nix/services/seaweedfs.nix
kashw2: Looks like there is one for filer: https://github.com/seaweedfs/seaweedfs/blob/5797fb24ec8974c7cd510225853069ce6b5d8bbd/weed/server/filer_server.go#L253

And S3 looks the same

test/flake.nix
kashw2: Linked to that PR in latest commit

nix/services/seaweedfs_test.nix
```diff
@@ -0,0 +1,27 @@
...
+      command = pkgs.writeShellApplication {
+        name = "seaweedfs-test";
+        runtimeInputs = [ pkgs.curl ];
+        text = ''
+          # The readiness probe already gates on all endpoints being up, so plain curls suffice.
+          curl -fsS -o /dev/null "http://${cfg.host}:${toString cfg.master.port}/cluster/healthz"
+          curl -fsS -o /dev/null "http://${cfg.host}:${toString cfg.volume.port}/healthz"
+          curl -fsS -o /dev/null "http://${cfg.host}:${toString cfg.filer.port}/"
+          curl -fsS -o /dev/null "http://${cfg.host}:${toString cfg.s3.port}/"
+        '';
```
shivaraj-bh: considering the test only runs if seaweedfs is healthy and it is only healthy if the health checks pass, should we have a better check or let the test be a no-op?

kashw2: A better test would probably involve having some sort of blob to upload. I don't mind removing the test / no oping it

shivaraj-bh: yes, we can noop it for now and improve it later.

kashw2: Not too sure if my implementation of the NOOP is what you have in mind, if not lmk and i'll amend it

shivaraj-bh: I made the change

shivaraj-bh:
Thanks for the contribution!

I didn’t let the CI run. aarch64-darwin is kinda borked, I need to fix it properly.

I ran the x86_64-linux CI locally.

PR #693, opened 2026-06-27
tempo: slim config; support v3;

kashw2:
This PR slims the config to just what is required for Tempo's HTTP backend to init and serve requests and also removes blocking and now deprecated config like the opencensus distributor from the generated tempo configuration which unblocks support for Tempo v3

### DIFF
--- a/nix/services/tempo.nix
+++ b/nix/services/tempo.nix
@@ -67,26 +67,6 @@ in
                     };
                   };
                 };
-                distributor = {
-                  receivers = {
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
I’d keep the diff to only the changes specific to the PR. could you revert the formatting changes?

kashw2:
> I’d keep the diff to only the changes specific to the PR. could you revert the formatting changes?

Fixed in latest commit

PR #700, opened 2026-07-15
feat: add `keycloak` service

gabyx:
I really like the services-flake and process-compose flake method.
**I was the maintainer for devenv.sh's keycloak service module** and 
I thought I upstream back a recent endeavor we took to make keycloak services-flake compatible:

- [x] Adapted from devenv.sh keycloak service which I contributed.
- [x] Added the basic tests.
- [x] Added documentation.
- [x] Added the tests with certificates.
- [x] Added `realm` export tests as well which were no more maintained in devenv. They however work.
- [x] Re-add realm export tests (lets see if they work, since the file-lock issue as TODO in devenv).

Assisted-by: no AI/Agent used.
Authored-by: @gabyx, @almutlue


gabyx:
@shivaraj-bh: Thanks for a review on this one. Happy to have that merged =).
I would close #656 in favor of this one.

nix/services/keycloak/service.nix
```diff
@@ -0,0 +1,243 @@
...
+      # Export all configured realms.
+      "${name}-realm-export-all" = lib.mkIf (realmsExport != [ ]) {
+        environment = keycloakEnv;
+        command = "${keycloak-realm-export-all}/bin/keycloak-realm-export-all";
+        disabled = true;
+        description = ''
+          Save the configured realms from keycloak, to back them up. You can run it manually.
+        '';
+      };
+    };
```
danbluhmhansen:
```suggestion
    settings.processes = {
      ${name} = {
        environment = keycloakEnv;
        command = "${lib.getExe keycloak-start}";
        readiness_probe = {
          exec = {
            command = "${lib.getExe keycloak-health}";
          };
          initial_delay_seconds = 10;
          timeout_seconds = 4;
          failure_threshold = 20;
        };
      };
    } // lib.optionalAttrs (realmsExport != [ ]) {
      # Export all configured realms.
      "${name}-realm-export-all" = {
        environment = keycloakEnv;
        command = "${keycloak-realm-export-all}/bin/keycloak-realm-export-all";
        disabled = true;
        description = ''
          Save the configured realms from keycloak, to back them up. You can run it manually.
        '';
      };
    };
```

The `lib.mkIf` approach does not seem to work for me, I get the following error, even using `exportRealms = false;`.

```
error: The option `perSystem.aarch64-darwin.process-compose.dev.settings.processes.kc1-realm-export-all.command' was accessed but has no value defined. Try setting the option.
```

shivaraj-bh:
Thanks! I will take a look. RE https://github.com/juspay/services-flake/pull/656. I did spend some time but I quickly realised that it required more work. On a quick glance this looks more or less what I was looking for in the former.

nix/services/keycloak/service.nix
gabyx: Thx, will have a look on that!

gabyx: @danbluhmhansen : Seems weird, why it really does not work, I replaced it with an `lib.optionalAttrs` append.
Do you might know why this `lib.mkIf` does not work?

danbluhmhansen: Hmm I am not an expert on `lib.mkIf` but I found this [mkIf vs if / then](https://discourse.nixos.org/t/mkif-vs-if-then/28521).

"the main difference is that mkIf produces an attribute set that is handled specially by the NixOS module system and keeps the contents available"

Maybe that has something to do with it.

gabyx: jeah, I know that it wraps it, but not sure, why it does not work on that attribute key in processes.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,298 @@
...
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config;
+
+  hasRealmExports = lib.any (lib.mapAttrsToList (realmName: opts: opts.export.enable) cfg.realms);
```
danbluhmhansen:
```suggestion
  hasRealmExports =
    cfg.exportRealms && lib.any (r: r.export.enable) (lib.attrValues cfg.realms);
```

`database.type = "dev-mem"` fails since `lib.any` takes two params it seems. I think we should also check `cfg.exportRealms` otherwise it gets ignored.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,298 @@
...
+          "dev-file"
+        ];
+        default = "dev-file";
+        example = "dev-mem";
+        apply =
+          val:
+          assert lib.assertMsg (val == "dev-mem" -> !hasRealmExports) ''
+            You cannot export realms with `realms.«name».export == true` when
+            using `database.type == 'dev-mem'`, import however works.
+            You can disable realms export with `exportRealms = true` globally.
```
danbluhmhansen:
```suggestion
            You can disable realms export with `exportRealms = false` globally.
```

Looks like this comment should be inverted.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,298 @@
...
+          absolute = false;
+        }
+      );
+      default = null;
+      example = "/run/keys/ssl_key";
+      description = ''
+        The path to a PEM formatted private key to use for TLS/SSL
+        connections.
+      '';
+    };
```
danbluhmhansen, line 52:
```suggestion
    sslCertificate = mkOption {
      type = types.nullOr (
        lib.types.pathWith {
          inStore = false;
        }
      );
      default = null;
      example = "/run/keys/ssl_cert";
      description = ''
        The path to a PEM formatted certificate to use for TLS/SSL
        connections.
      '';
    };

    sslCertificateKey = mkOption {
      type = types.nullOr (
        lib.types.pathWith {
          inStore = false;
        }
      );
      default = null;
      example = "/run/keys/ssl_key";
      description = ''
        The path to a PEM formatted private key to use for TLS/SSL
        connections.
      '';
    };
```

Examples use absolute paths, so I think we should remove `absolute = false;`.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,298 @@
...
+              example = "./realms/a.json";
+              description = ''
+                The path (relative to the `process-compose` working dir or an Nix store path)
+                where you want to import (or export) this realm «name» to.
+                - If not set and `import` is `true` this realm is not imported.
+                - If
+                  - set to an Nix store path
+                  - or not it is not set
+                  and `export` is `true` then
+                  it is exported to `''${config.dataDir}/realm-export/«name».json`.
```
danbluhmhansen, line 145:
```suggestion
                The path (relative to the `process-compose` working dir or an Nix store path)
                where you want to import (or export) this realm «name» to.
                - If not set this realm is not imported.
                - If
                  - set to an Nix store path
                  - or it is not set
                  and `export.enable` is `true` then
                  it is exported to `''${config.dataDir}/realm-export/«name».json`.
```

Example does not match changes from the devenv schema.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,298 @@
...
+                    absolute = false;
+                  }
+                );
+                default = null;
+                example = "./realms/a.json";
+                description = ''
+                  The path (relative to the `process-compose` working dir or an Nix store path)
+                  where you want to export this realm «name» to.
+                  - If not set and `import.path` is relative it is used as default.
+                  - If not set otherwise its defaulted to a value in the data folder.
```
danbluhmhansen, line 167:
```suggestion
                  The path (relative to the `process-compose` working dir or an Nix store path)
                  where you want to export this realm «name» to.
                  - If not set and `import` is relative it is used as default.
                  - If not set otherwise its defaulted to a value in the data folder.
```

Example does not match changes from the devenv schema.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,298 @@
...
+
+      example = lib.literalExpression ''
+        {
+          myrealm = {
+            path = "./myfolder/export.json";
+            import = true; # default
+            export = true;
+          };
+        }
+      '';
```
danbluhmhansen, line 192:
```suggestion
      example = lib.literalExpression ''
        {
          myrealm = {
            import = "./myfolder/realm.json";
            export = {
              enable = true;
              path = "./myfolder/export.json";
            };
          };
        }
      '';
```

Example does not match changes from the devenv schema.

nix/services/keycloak/service.nix
```diff
@@ -0,0 +1,246 @@
...
+
+  keycloakEnv = {
+    # Note: we add "./" cause keycloak's database url
+    # does not allow implicitly relative paths.
+    KC_HOME_DIR = "./" + cfg.dataDir;
+    KC_CONF_DIR = cfg.dataDir + "/conf";
+    KC_TMP_DIR = cfg.dataDir + "/tmp";
+
+    KC_BOOTSTRAP_ADMIN_USERNAME = "admin";
+    KC_BOOTSTRAP_ADMIN_PASSWORD = "${lib.escapeShellArg cfg.initialAdminPassword}";
```
danbluhmhansen:
```suggestion
    KC_BOOTSTRAP_ADMIN_PASSWORD = cfg.initialAdminPassword;
```

I think we do not need to use escapeShellArg here since it goes to the process-compose env vars.

nix/services/keycloak/service.nix
```diff
@@ -0,0 +1,246 @@
...
+        ${keycloakBuild}/bin/kc.sh show-config || true
+
+        echo "Import realms (if any)..."
+        ${builtins.concatStringsSep "\n" realmImport}
+        echo "========================"
+
+        echo "Start keycloak:"
+        exec ${keycloakBuild}/bin/kc.sh start --optimized --import-realm
+      '';
+in
```
danbluhmhansen, line 198:
```suggestion
in
assert lib.assertMsg ((cfg.sslCertificate == null) == (cfg.sslCertificateKey == null))
  "Both `sslCertificate` and `sslCertificateKey` must be set together, or neither. Got only one.";
```

If either ssl cert options are null we silently fall back to dummy certs. Would be nice with a check.

nix/services/keycloak/test-realms/export.nix
```diff
@@ -0,0 +1,84 @@
...
+          return 1
+        fi
+
+        if [ ! -f "${realmDstDir}/master.json" ]; then
+          echo "!! Realm '${realmDstDir}/master.json' did not get exported."
+          return 1
+        fi
+
+        if [ ! -f "${realmDstDir}/test.json" ]; then
+          echo "!! Realm '${realmDstDir}/test.json'  did not get exported".
```
danbluhmhansen:
```suggestion
          echo "!! Realm '${realmDstDir}/test.json' did not get exported."
          return 1
```

Missing return, and minor stray `.`.

nix/services/keycloak/service.nix
```diff
@@ -0,0 +1,246 @@
...
+        if builtins.isInt v then
+          toString v
+        else if builtins.isString v then
+          v
+        else if true == v then
+          "true"
+        else if false == v then
+          "false"
+        else if isSecret v then
+          builtins.hashString "sha256" v._secret
```
danbluhmhansen: I am not sure I get this 100%, but it looks like paths provided here using `_secret` are just hashed, but does not actually perform any replacements at runtime? Easiest fix would be to just remove this option for now, and then also remove the documentation of this feature in options.nix.

danbluhmhansen (changes requested):

nix/services/keycloak/options.nix
gabyx: Thanks! Good catch!

nix/services/keycloak/options.nix
gabyx: Correct.  A bit unsure however why one wants to pass a an absolute path. Shouldn't we always take relative paths?
I relaxed it and added a note that relative paths are relative to working dir.

nix/services/keycloak/service.nix
gabyx: jeah, true, but now looking at it: guess its a relict from devenv.sh and what they did.

nix/services/keycloak/options.nix
gabyx: Thx!

nix/services/keycloak/service.nix
gabyx: Nice catch!

nix/services/keycloak/service.nix
gabyx: Oh good catch, thats a relict from the devenv stuff. I removed it.

gabyx:
@danbluhmhansen: Thanks for the nice review, catched a lot of iteration mistakes still left in! :love_letter:

gabyx:
@danbluhmhansen: I needed to squash all review changes into one last commit for you to re-review.
I got an infinte recursion on the lib.assert, which I moved.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,315 @@
...
+
+    settings = mkOption {
+      type = lib.types.submodule {
+        freeformType = types.attrsOf (
+          types.nullOr (
+            types.oneOf [
+              types.str
+              types.int
+              types.bool
+              (types.attrsOf types.path)
```
danbluhmhansen, line 208:
```suggestion
```

Nitpick. Only used for the `_secret` options, could be removed if we like.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,315 @@
...
+            '';
+          };
+        };
+      };
+
+      example = lib.literalExpression ''
+        {
+          hostname = "localhost";
+          https-key-store-file = "/path/to/file";
+          https-key-store-password = { _secret = "/run/keys/store_password"; };
```
danbluhmhansen:
```suggestion
          https-key-store-password = "change_me";
```

Nitpick. Example no longer supported.

nix/services/keycloak/options.nix
```diff
@@ -0,0 +1,315 @@
...
+
+        Options containing secret data should be set to an attribute
+        set containing the attribute `_secret` - a
+        string pointing to a file containing the value the option
+        should be set to. See the example to get a better picture of
+        this: in the resulting
+        {file}`conf/keycloak.conf` file, the
+        `https-key-store-password` key will be set
+        to the contents of the
+        {file}`/run/keys/store_password` file.
```
danbluhmhansen:
```suggestion
```

Nitpick. Docs suggest using the _secret options but that will now throw.

danbluhmhansen (approved):
Just minor suggestions left from me, otherwise looks good!

nix/services/keycloak/options.nix
gabyx: I do not get that suggestion: isnt it that the

```nix
freeformType = types.attrsOf (
          types.nullOr (
            types.oneOf [
              types.str
              types.int
              types.bool
              (types.attrsOf types.path)
            ]
          )
        );
```
Only allows these types when adding other non-modules keycloak options?
Why do we only want `types.path`?`

nix/services/keycloak/options.nix
gabyx: ah good catch I removed that stuff completelyy.

nix/services/keycloak/options.nix
gabyx: Ah, ok since I removed the _secrets thing. But I still leave it in, I think its find to specify `types.path`.

PR #701, opened 2026-07-15
feat: add `.editorconfig` file

gabyx:
- Add a `.editorconfig` file for better editor setup and indentation (std best-practice).

To not repeat my self: https://editorconfig.org/#overview.
It helps with different IDE setups and became a defacto std. as all editors implement that, nvim, vscode, zed etc.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# EditorConfig is awesome: https://EditorConfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
I haven’t used `.editorconfig` before. Could you add a short note about the problem it solved for you (or in this case, will solve, as it is still unmerged)? that’d make a case for merging this for other contributors too, aside from being a std best practice.

gabyx:
Added an info above.

PR #703, opened 2026-07-16
feat: add `authentik` service

gabyx:
I tried to configure an `authentik` service with process-compose based on the packages in https://github.com/nix-community/authentik-nix. I was quite successful I think.

- [x] Documentation
- [x] Tests  health point + blueprint export)
- [x] Export the blueprints (db dumb etc) with `blueprints.export.enable = true` and triggering the process.
- [x] The modules does not pull in the `authentic-nix` flake but requests it on `components` option.
- [x] `authentik` service needs a sidecar services `postgres`. 
  These are easily available configured already under `services.postgres`. See the documentation and the tests.
  *Note*: This implementation actually posed the question how an `authentik` module can configure another service as a sidecar. Currently how the modules are loaded we have no access to the overall toplevel options. @shivaraj-bh (maybe there is a better option here). Thats why I went with this solution.

Note: The tests need `authentik-nix` input and a small overlay such that `pkgs.authenticComponents` (overrideScope) is available.

**WARNING**: Should be merged when upstream https://github.com/nix-community/authentik-nix/pull/156 is merged.

Assited-by: Claude Opus 4.8 **only** for debugging and design decisions on this module. (Documentation is not super great in authentik)


gabyx:
@srid: Would be cool if somebody could review this =). Its already used and works well.

PR #706, opened 2026-07-22
fix: add `rustfs` service

gabyx:
Add the `rustfs` service, implementation adapted from devenv.

- [x] Documentation
- [x] Tests:  health point, import/export of iam-settings, bucket creation.
- [x] The modules does not pull in the `rustfs` flake but requests it on `package` option.

Note: The tests need `rsutfs` input and a small overlay such that `pkgs.rustfs` is available.

Assisted-by: Claude Opus 4.8 **only** for documentation ".md" generation -> reviewed and read by me.


gabyx:
@srid: Can I request maybe a review on that one. This service is quite simple.

srid:
@shivaraj-bh Could you take a look?

PR #708, opened 2026-07-25
feat: add `treefmt-nix` to dev flake

### LINKED ISSUE #707: fix: add treewide formatting with `nixfmt` and others
@shivaraj-bh:

May I suggest a small PR with a `treefmt-nix` setup which uses `nixfmt` + some others like `prettier` for `.md` files etc to format treewide everything, such that 
other contributors have less pain in their PRs.

- `just format`
- `format` check in `omnix.nix`

Also `nixpkgs-fmt` is not deterministic enough which is a pain at the moment in the repository, a small line change still might go through the formatter 
succesfully, so it is **too lenient** + outdated.

gabyx:
## Things Done

Closes: #707 

- [x] Added `treefmt` setup in `dev` flake to format treewide all files.
- [x] Needed to update `nixpkgs` since `treefmt-nix` needs a newer input.
- [x] Flake check in `dev` with `nix flake check` runs the formatting check. (Should be integrated in omnix by default?)
- [x] Replaced `nixpkgs-fmt` with `treefmt` on `pre-commit`.
- [x] `.prettierrc.yaml` in root (could be maybe somewhere else ...) to format all markdown files consistently. 

## How it works: 
- [x] `just format` will format all files: Once happy with the setup one can format the files on `main` in one commit. The `docs` folder with the markdown files look well treated by prettier and there is not much to change I guess.

Note: When merging, one should apply `just format` on main directly afterwards as `chore: formatting all files treewide`

Assisted-by: no agent.


gabyx:
@shivaraj-bh: Could you maybe run `just format` on that branch and check how it looks?

shivaraj-bh:
@gabyx Why use `treefmt-nix` when pre-commit has a nixfmt hook?

gabyx:
@shivaraj-bh : Cause running hooks and formatting are two orthogonal things. Running hooks is fine but having one global process which formats treewide (with caching) is something which is not super easy achievable with hooks.

Running the format step (the exact same treefmt build and not again another nixfmt hook) in pre-commit is however the better option IMO.
Also you dont need to worry about other formats, its covered all.

Treefmt is awesome btw :). Once used you will probably deploy it everywhere :) (We did so and its a game changer :))

Malix-Labs (approved):

shivaraj-bh:
> Running hooks is fine but having one global process which formats treewide (with caching) is something which is not super easy achievable with hooks.

I run `pre-commit run nixpkgs-fmt` which formats treewide. It doesn’t cache though. Since speed was never (not yet) really a problem I didn’t bother switching.

> Treefmt is awesome btw :). Once used you will probably deploy it everywhere :) (We did so and its a game changer :))

I agree. I use it in multiple projects. I don’t mind adding it here.

dev/nix/treefmt.nix
```diff
@@ -0,0 +1,59 @@
...
+          ];
+        };
+
+        # Markdown, JSON, YAML, etc.
+        programs.prettier.enable = true;
+        settings.formatter.prettier.excludes = [ ];
+
+        # Nix.
+        programs.deadnix.enable = false;
+        programs.statix.enable = false;
```
shivaraj-bh: aren’t they disabled by default?

justfile
```diff
@@ -1,3 +1,6 @@
+set positional-arguments
+set shell := ["bash", "-cue"]
+
```
shivaraj-bh: I haven’t used them before. Is this really needed?

justfile
```diff
@@ -6,6 +9,13 @@ default:
 changelog:
     cz ch --start-rev $(git describe --tags --abbrev=0 HEAD^) --incremental
 
+# Format treewide all files.
+[group('lint')]
+format *args:
+    cd dev && nix run \
+        --override-input services-flake ../. \
+        ".#treefmt" -- "$@"
```
shivaraj-bh: we can avoid `nix run`’ing by including treefmt package in the dev flake’s devshell output

shivaraj-bh:
> @shivaraj-bh: Could you maybe run `just format` on that branch and check how it looks?

@gabyx looks good. ~I will merge this and then format in a different PR.~

EDIT: or maybe just another commit in the same PR and then “Rebase and merge"

dev/nix/treefmt.nix
gabyx: Sorry, was more a suggestion -> good point. Should either enable them or remove them. 
What would you like?

justfile
gabyx: Yes, generally recommended for making  `"$@"`  correctly work AFAIK.
I would set `shell` as well cause its explicit what you expect in your inline scripts.
Otherwise all bets are off and you basically say POSIX `sh`, I dont think you want that or do you?

justfile
gabyx: Jeah thats true! Lets do it.

dev/nix/treefmt.nix
shivaraj-bh: While they are definitely useful. I’d keep this PR focused on formatting.

justfile
shivaraj-bh: maybe a comment will be useful for when I (or future maintainers) read it at some point

dev/nix/treefmt.nix
gabyx: Will remove it.

justfile
gabyx: I can add it. yes.

gabyx:
@srid: Last commit contains all changes for an inspection.

shivaraj-bh:
Thanks @gabyx !

Let me know if it is good for me to take over. I will squash the existing commits into one; add a treewide format commit; run CI; Rebase and merge.

gabyx:
@srid: Yes, its good, to take over, thanks a lot.

gabyx:
@srid: maybe some precaution, maybe check the documentation is not destroyed, since markdown files are formatted, if yes -> let me know -> set a global ignore until the markdown is fixed. It looked good, but just saying.

shivaraj-bh:
I have split the commits to make it easier for review. I have also changed few things in https://github.com/juspay/services-flake/pull/708/commits/a51dec282c85800afccba08d356170e13b12e36a, notably:
- removed `build.check` as it was redundant since pre-commit already has a flake check that uses treefmt to check formatting
- removed `set shell` from justfile. To keep only formatting related changes in this PR

PR #716, opened 2026-08-11
Add `live_store` configuration for Tempo v3

kashw2:
Tempo v3 added `live_store` as a configurable field which has two directory path sub-fields. Consumers of this module currently have to set this themselves if they are using a version of nixpkgs that ships v3. This pull request resolves that issue by adding a version guard that conditionally adds the `live_store` block for any consumers of this module using Tempo v3

### DIFF
--- a/nix/services/tempo.nix
+++ b/nix/services/tempo.nix
@@ -65,6 +65,13 @@ in
                     local = {
                       path = "${config.dataDir}/blocks";
                     };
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
could you revert the formatting changes? formatting will be standardised after https://github.com/juspay/services-flake/issues/707

PR #717, opened 2026-08-15
Grafana Alloy Init

kashw2:
Grafana Alloy is OTEL compliant telemetry collector and pipelinable distributor

https://grafana.com/docs/alloy/latest/
https://github.com/grafana/alloy

### DIFF
--- a/nix/services/alloy_test.nix
+++ b/nix/services/alloy_test.nix
@@ -1,0 +1,20 @@
+{
+  pkgs,
+  lib,
+  ...
+}:
+{
+  services.alloy."alloy1" = {
+    enable = true;
+    configFile = pkgs.writeText "config.alloy" ''
+      livedebugging {
+        enabled = true
+      }
+    '';
+  };
+
+  settings.processes.test = {
+    command = lib.getExe' pkgs.coreutils "true";
+    depends_on."alloy1".condition = "process_healthy";
+  };
+}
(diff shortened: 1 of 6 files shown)


nix/services/alloy_test.nix
```diff
@@ -0,0 +1,27 @@
...
+    let
+      cfg = config.services.alloy."a1";
+    in
+    {
+      command = pkgs.writeShellApplication {
+        name = "alloy-test";
+        runtimeInputs = [ pkgs.curl ];
+        text = ''
+          curl -sSfN "http://${cfg.listenAddress}:${toString cfg.port}/-/ready"
+        '';
```
shivaraj-bh: This is a no-op since the `depends_on` below already does the same check. you can take a look at seaweedfs_test.nix for what to do in the case of a no-op.

kashw2: Fixed in the latest commit, the command now mimics what happens in seaweedfs_text.nix.

Thanks for the review

shivaraj-bh:
needs a `just fmt`

nix/services/alloy_test.nix
```diff
@@ -0,0 +1,20 @@
...
+    configFile = pkgs.writeText "config.alloy" ''
+      livedebugging {
+        enabled = true
+      }
+    '';
+  };
+
+  settings.processes.test = {
+    command = lib.getExe' pkgs.coreutils "true";
+    depends_on."a1".condition = "process_healthy";
```
shivaraj-bh:
```suggestion
    depends_on.”alloy1".condition = "process_healthy";
```

sorry, missed this in my previous review

kashw2: No worries, my mistake

